PR #125, opened 2026-03-09
feat: return error when there is no root project without `--all-projects`

thomasschafer:
Currently we have a bug where, when testing or monitoring a uv virtual workspace (i.e. a workspace with no root package, only workspace packages), all workspace packages will be included in the results. This is wrong, as a virtual workspace has no root and hence no dependencies. This PR adds an error in this scenario, informing the user that they can instead use `--all-projects` if they would like to intentionally include workspace packages in the results.

I also updated some fixtures in other tests to include the `is_project_root` property, and add some other missing data which ensures that they are as realistic as possible.

### DIFF
--- a/internal/uv/uv_plugin.go
+++ b/internal/uv/uv_plugin.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/snyk/dep-graph/go/pkg/depgraph"
 
+	"github.com/snyk/error-catalog-golang-public/opensource/ecosystems"
+
 	"github.com/snyk/cli-extension-dep-graph/internal/conversion"
 	"github.com/snyk/cli-extension-dep-graph/internal/snykclient"
 	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems/discovery"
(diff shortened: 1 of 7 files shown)


internal/uv/uv_plugin.go
```diff
@@ -86,13 +88,25 @@ func (p Plugin) buildFindings(
...
 ) ([]scaplugin.Finding, error) {
 	parsedSbom, err := parseAndValidateSBOM(sbom)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse and validate sbom for %s: %w", lockFilePath, err)
 	}
 
+	if !options.AllProjects && !options.UvWorkspacePackages && !hasProjectRoot(parsedSbom) {
+		log.Info(ctx, "No root project found in SBOM", logger.Attr("lockFile", lockFilePath))
+		noRootErr := ecosystems.NewUvNoProjectRootError(
+			"Found uv project with no root. To scan non-root projects use the --all-projects flag.",
```
snyk-tim: I think we should stick to the same terminology that Astral use when warning the user. As this is related to virtual workspaces it would be better if the message was easier for the end user to understand. So something like:

```
Found uv workspace with no root project. To scan all workspace members use the --all-projects flag.
```

thomasschafer: Good point, done!

snyk-tim (approved):

PR #129, opened 2026-03-17
Fix/pip error in cause

james-snyk:

### What this does

Store full error in causes so we can return a nice error to the user via the details but have extra info to use to decide how to process the given error.


pkg/ecosystems/python/pip/report.go
```diff
@@ -230,7 +230,7 @@ func classifyPipError(ctx context.Context, log logger.Logger, err error) error {
 		strings.Contains(stderr, "Invalid version") {
 		return ecosystems.NewSyntaxIssuesError(
 			fmt.Sprintf("Invalid syntax in requirements file: %s", stderr),
-			snyk_errors.WithCause(pipErr.err),
+			snyk_errors.WithCause(errors.New(stderr)),
```
james-snyk, line 233: pipErr.err was always just exit code 1 so not very helpful

calhar-snyk (approved):

PR #134, opened 2026-03-31
chore: align uv plugin to canonical interface

thomasschafer:
- [ ] Tests written and linted [ℹ︎](https://github.com/snyk/general/wiki/Tests) - n/a
- [ ] Documentation written [ℹ︎](https://github.com/snyk/general/wiki/Documentation) - n/a

### What this does

While working on uv we have had a separate interface to the canonical interface used by the pip plugin. Now that the uv work has stabilised and we are unlikely to need to change the interface very much, we can remove the interface used by uv and move across to the canonical interface. This interface needed adjusting slightly to account for the additional fields required by uv.

I also tidied up the doc at `pkg/ecosystems/README.md` as some of it was out of date e.g. on the interface definition, and just pointing directly at the relevant code seemed cleaner than having duplication across code and docs.

I would highly recommend reviewing each commit individually as otherwise the diff will be very confusing.

**Note**: The integration tests are failing because they break every time new versions of transitive dependencies are released. I have a fix for that [here](https://github.com/snyk/cli-extension-dep-graph/pull/137/) which I'll merge before this PR.


internal/conversion/sbom.go
```diff
@@ -9,13 +9,18 @@ import (
 
 	"github.com/snyk/cli-extension-dep-graph/internal/snykclient"
 	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems/logger"
-	"github.com/snyk/cli-extension-dep-graph/pkg/scaplugin"
 )
 
+type Metadata struct {
+	PackageManager string
+	Name           string
+	Version        string
```
mcombuechen: Sorry if this seems like an ignorant question, but what is Metadata, Name and Version in this case? Should we document the fields briefly?

thomasschafer: Just moved from [here](https://github.com/snyk/cli-extension-dep-graph/blob/852dbfb/pkg/scaplugin/interface.go#L22) but yes I'll document

thomasschafer: Done!

mcombuechen (approved):

james-snyk (approved):

PR #137, opened 2026-04-02
test: stop asserting exact versions for unpinned pip deps

thomasschafer:

### What this does

The Python integration tests were failing whenever a transitive pip dependency released a new patch version. For instance, `charset-normalizer` going from `3.4.6` to `3.4.7` broke every fixture that mentioned it (e.g. in [this PR](https://github.com/snyk/cli-extension-dep-graph/pull/134/) which didn't make any relevant changes to pip processing).

This PR fixes that by making the following changes:

- Fixture files now use `"*"` as the version for packages not explicitly pinned in `requirements.txt`, i.e. transitive deps and direct deps specified with range constraints (e.g. `requests>=2.28.0`). Pinned packages like `requests==2.31.0` or `black@23.12.1` (via git ref) keep their exact versions.
- A new `resolveWildcardVersions` helper in the test substitutes `"*"` wildcard versions in the expected fixture with the actual resolved versions before comparing, so the structural assertion still holds without caring which patch release pip chose.
- Per-Python-version fixture files (`expected_plugin_3.8.json` etc.) are kept only where the graph structure differs between Python versions:
  - `env-markers`: has `python_version < "3.10"` markers, so the package set changes
  - `git-references`: black needs `tomli`/`typing-extensions` on Python < 3.11
  - `multi-requirements`: flask/click pull in `importlib-metadata` on Python < 3.10
  - Files that differed only in transitive dependency versions are deleted.
- `tools/update-fixtures/` and its Makefile targets are removed, as they are no longer needed.

Fixtures will now only need updating if the structural shape of a dependency graph changes — a new transitive dep is introduced, an existing one is removed, or a pinned direct dep version changes in `requirements.txt`.


37IulianPopovici (approved):
LGTM

can you please reply to the PR Review Guide answering why its comment is not applicable in this example? 🙏

thomasschafer:
@snyk-pr-review-bot regarding this comment:

> The extractNameVersionMap function uses strings.LastIndex(id, "@") to separate a package name from its version. However, Python dependencies frequently use git URLs or references containing '@' characters (e.g., pkg @ git+https://.../repo.git@v1.0). In these cases, LastIndex will incorrectly split the string, leading to an incorrect name key in the map and causing substituteWildcards to fail to resolve the intended package versions.

We don't use any such dependencies in these tests so this would be an unnecessary optimisation at this stage. If in future we use these we can work out a proper solution.

thomasschafer:
> LGTM
> 
> can you please reply to the PR Review Guide answering why its comment is not applicable in this example? 🙏

Thanks @37IulianPopovici, done!

PR #139, opened 2026-04-10
feat!: swap metadata with ProjectDescriptor

james-snyk:

### What this does

Swap metadata with ProjectDescriptor what reports information about the given project

### DIFF
--- a/pkg/identity/project.go
+++ b/pkg/identity/project.go
@@ -1,0 +1,28 @@
+package identity
+
+// ProjectDescriptor contains the complete description of a project,
+// including its identity and build arguments.
+type ProjectDescriptor struct {
+	Identity ProjectIdentity
+}
+
+// ProjectIdentity defines the core identifying characteristics of a project.
+type ProjectIdentity struct {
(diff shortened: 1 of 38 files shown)


pkg/project/project.go
```diff
@@ -0,0 +1,32 @@
+package project
```
mcombuechen: should we call this package `identity`?

pkg/identity/project.go
```diff
@@ -0,0 +1,32 @@
+package project
+
+// ProjectDescriptor contains the complete description of a project,
+// including its identity and build arguments.
+//
+//nolint:revive // Keeping ProjectDescriptor name for compatibility
+type ProjectDescriptor struct {
+	Identity ProjectIdentity
+}
```
mcombuechen, line 7: Should we add build args for the sake of completeness, also so we can use this in SCM?

james-snyk: The build args are legacy and with this new flow we can move away from them.

JamesPatrickGill (approved):

mcombuechen (approved):

PR #140, opened 2026-04-10
feat: add type to project identifiers

mcombuechen:
### What this does

Adds a `Type` to the Project Identifier values; extracted from the dep-graph.

### DIFF
--- a/pkg/ecosystems/orchestrator/orchestrator.go
+++ b/pkg/ecosystems/orchestrator/orchestrator.go
@@ -7,6 +7,12 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/rs/zerolog"
+	"github.com/snyk/dep-graph/go/pkg/depgraph"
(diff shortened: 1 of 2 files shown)


pkg/ecosystems/orchestrator/orchestrator.go
```diff
@@ -224,8 +227,9 @@ func depGraphOutputToSCAResult(output *parsers.DepGraphOutput, ictx workflow.Inv
 
 		result.DepGraph = &dg
 
-		// Extract runtime from depgraph if available
 		if dg.PkgManager.Name != "" {
+			result.ProjectDescriptor.Identity.Type = dg.PkgManager.Name
+			// Extract runtime from depgraph if available
 			result.ProjectDescriptor.Identity.TargetRuntime = &dg.PkgManager.Name
```
mcombuechen, line 233: This looks like a placeholder to me... why would we use the package manager's name?

james-snyk (approved):

PR #141, opened 2026-04-14
feat: set target file for identity

james-snyk:

### What this does

Set target file for identity based on ecosystem.

Add --project-name flag and set pip/pipenv root to be equal to the name of the directory where the file is located just like the original python plugin.

### DIFF
--- a/pkg/ecosystems/orchestrator/orchestrator.go
+++ b/pkg/ecosystems/orchestrator/orchestrator.go
@@ -236,3 +236,29 @@ func depGraphOutputToSCAResult(output *parsers.DepGraphOutput, ictx workflow.Inv
 
 	return result, nil
 }
+
+// getProjectTargetFileBasedOnType determines if a plugin sets targetFile based on package manager type.
+// Plugin audit: https://snyksec.atlassian.net/wiki/spaces/TOE1/pages/4651909131/Plugin+Audit
+func getProjectTargetFileBasedOnType(projectType, file string) *string {
+	switch projectType {
+	case "pip":
+		// snyk-python-plugin sets targetFile for Pipfile and setup.py, but not for requirements.txt
+		if strings.HasSuffix(file, "requirements.txt") {
+			return nil
+		}
+		return &file
+	case "gradle":
+		// snyk-gradle-plugin sets targetFile for build.gradle.kts but not for build.gradle
+		if strings.HasSuffix(file, "build.gradle.kts") {
+			return &file
+		}
+		return nil
+	case "poetry", "gomodules", "golangdep", "nuget", "paket", "composer", "cocoapods", "hex", "swift":
+		// These plugins set relative path to provided targetFile
+		return &file
+	case "npm", "yarn", "pnpm", "maven", "sbt", "rubygems":
+		// These plugins do not set plugin.targetFile
+		return nil
+	}
+	return nil
+}
--- a/pkg/ecosystems/orchestrator/orchestrator.go
+++ b/pkg/ecosystems/orchestrator/orchestrator.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/rs/zerolog"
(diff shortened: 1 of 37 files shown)


pkg/ecosystems/orchestrator/orchestrator_test.go
```diff
@@ -14,66 +14,170 @@ import (
...
+			require.Len(t, results, 1)
+
+			assert.Equal(t, test.expectedType, results[0].ProjectDescriptor.Identity.Type)
+			assert.Equal(t, test.expectedTargetFile, results[0].ProjectDescriptor.GetTargetFile())
+			if test.expectedTargetRuntime != nil {
+				require.NotNil(t, results[0].ProjectDescriptor.Identity.TargetRuntime)
+				assert.Equal(t, *test.expectedTargetRuntime, *results[0].ProjectDescriptor.Identity.TargetRuntime)
+			} else {
+				assert.Nil(t, results[0].ProjectDescriptor.Identity.TargetRuntime)
+			}
```
mcombuechen: Can we keep the TargetRuntime part separate? I think the tabular tests are great for the TargetFile part, but doing a roundhouse kick and including the TargetRuntime and Type as well just increases complexity in these tests and makes them harder to maintain.

james-snyk: I can change it back i only did it this way due to the tests the selves being identical just checking a different field.

mcombuechen: Yeah I know I'm opinionated with my comment, and there's a lot of repeat code. In tests I feel like you actually sacrifice readability when you make them DRYer. One distinct test for just one thing reads much better to me than a complex one which tries to avoid a repeat setup.

james-snyk: I moved the common part of the test into a helper function instead and split the tests back up.

mcombuechen (approved):

pkg/ecosystems/python/pip/depgraph_test.go
```diff
@@ -73,7 +73,7 @@ func TestReport_ToDependencyGraph(t *testing.T) {
 			},
 		}
 
-		dg, err := report.ToDependencyGraph(context.Background(), logger.Nop(), PkgManagerPip)
+		dg, err := report.ToDependencyGraph(context.Background(), logger.Nop(), PkgManagerPip, "")
```
tonidevine1, line 76: All these tests pass an empty string for project name, which defaults to root. Is it worth adding a test that passes a real project name?

pkg/ecosystems/orchestrator/orchestrator.go
```diff
@@ -229,10 +229,41 @@ func depGraphOutputToSCAResult(output *parsers.DepGraphOutput, ictx workflow.Inv
...
 
 	return result, nil
 }
+
+// getProjectTargetFileBasedOnType determines if a plugin sets targetFile based on package manager type.
+func getProjectTargetFileBasedOnType(projectType, file string, isWorkspace bool) *string {
+	switch projectType {
+	case "pip":
+		// snyk-python-plugin sets targetFile for Pipfile and setup.py, but not for requirements.txt
+		if strings.HasSuffix(file, "requirements.txt") {
```
tonidevine1: Maybe more for my understanding but do we want to use HasSuffix for identifying manifests? So we'd allow anything ending with that suffix.

tonidevine1:
Also are the AI comments talking rubbish here?

pkg/ecosystems/python/pip/depgraph_test.go
james-snyk: The integration tests  with real projects and makes sure the project name is set based on the project path.

pkg/ecosystems/orchestrator/orchestrator.go
james-snyk: We will always get a relative path back from the cli and from looking at the legacy cli the only thing it does is make sure it doesnt end with requirements.txt. We could maybe make use of the paths package to get the file name

tonidevine1: Great, thanks for clearing this up!

tonidevine1 (approved):

PR #144, opened 2026-04-16
feat: add new --allow-incomplete-sbom-flag [CSENG-175]

snyk-abedonik:
## Checklist

- [x] Tests written and linted
- [ ] Documentation written
- [x] Commit history is tidy

---

## What this does

When running `snyk sbom --all-projects`, the CLI previously used `fail-fast` mode: if **any** project in the workspace failed to resolve (e.g. missing lockfile, unsupported manifest), the entire SBOM generation was aborted and no output was produced.

This PR introduces the `--allow-incomplete-sbom` flag that lets users opt in to a **partial-success** mode. When the flag is set alongside `--all-projects`:

- Projects that resolve successfully are included in the generated SBOM.
- Projects that fail are collected as structured `ScanError` entries (with a `subject` path and human-readable `text` message) and forwarded to the SBOM service alongside the successful dep-graphs.

## References
https://docs.google.com/document/d/1vhRKlienHz1kbrCI-2BJ3maO6ykmlAz-hSApgo8MGEw/edit
https://docs.google.com/document/d/1i4exfAq3Dvoy_mKwQAwL3LYE6_Qkt7jQVYVOSzijZdw/edit
https://docs.google.com/document/d/1j0gNbzCALFF3WfIxLd5PVBtglJb4kYGQdheoM27VMaY/edit

### DIFF
--- a/pkg/depgraph/legacy_resolution.go
+++ b/pkg/depgraph/legacy_resolution.go
@@ -12,6 +12,11 @@ import (
 	"github.com/snyk/cli-extension-dep-graph/pkg/depgraph/parsers"
 )
 
+const (
+	printGraphCliArg                 = "--print-graph"
+	printOutputJsonlWithErrorsCliArg = "--print-output-jsonl-with-errors"
+)
+
 var legacyWorkflowID = workflow.NewWorkflowIdentifier(legacyCLIWorkflowIDStr)
 
 func handleLegacyResolution(ctx workflow.InvocationContext, config configuration.Configuration, logger *zerolog.Logger) ([]workflow.Data, error) {
@@ -53,7 +58,11 @@ func chooseGraphArgument(config configuration.Configuration) (string, parsers.Ou
 		return "--print-effective-graph-with-errors", parsers.NewJSONL()
 	}
 
-	return "--print-graph", parsers.NewPlainText()
+	if config.GetBool(FlagPrintOutputJsonlWithErrors) {
+		return printOutputJsonlWithErrorsCliArg, parsers.NewJSONL()
+	}
+
+	return printGraphCliArg, parsers.NewPlainText()
 }
 
 func mapToWorkflowData(depGraphs []parsers.DepGraphOutput, logger *zerolog.Logger) []workflow.Data {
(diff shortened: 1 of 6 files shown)


pkg/depgraph/legacy_resolution.go
```diff
@@ -53,7 +58,11 @@ func chooseGraphArgument(config configuration.Configuration) (string, parsers.Ou
 		return "--print-effective-graph-with-errors", parsers.NewJSONL()
 	}
 
-	return "--print-graph", parsers.NewPlainText()
+	if config.GetBool(FlagPrintOutputJsonlWithErrors) {
+		return printOutputJsonlWithErrorsCliArg, parsers.NewJSONL()
+	}
```
snyk-tim, line 63: Passing `--print-output-jsonl-with-errors` instead of `--print-graph` to the legacy cli via `snyk test` can result in a different SBOM being produced for the same project.

For instance, there are at least two ecosystem plugins that I am aware of that have a side-effect when `--print-graph` is passed as an argument:

- `snyk-maven-plugin` uses `--print-graph` to set the `verbose` [variable](https://github.com/snyk/snyk-mvn-plugin/blob/master/lib/index.ts#L136-L139) in the Maven plugin, which results in the following:
  - Calls the `mvn` cli with the `-Dverbose` flag [here](https://github.com/snyk/snyk-mvn-plugin/blob/2d0e29b7c4a580e57ef8e74398ebf0f308e85370/lib/maven/dependency-tree.ts#L64-L70) which results in all dependency candidates considered during resolution being returned.
  - It will also [prune the depgraph](https://github.com/snyk/snyk-mvn-plugin/blob/2d0e29b7c4a580e57ef8e74398ebf0f308e85370/lib/parse/dep-graph.ts#L22-L25).
- `snyk-gradle-plugin` uses `--print-graph` to set the `verbose` [variable](https://github.com/snyk/snyk-gradle-plugin/blob/main/lib/index.ts#L494) in the Gradle plugin, which results in the following:
  - It will [prune the depgraph](https://github.com/snyk/snyk-gradle-plugin/blob/main/lib/graph.ts#L107-L116)

Ideally the long term goal will be to improve this. A work around could be to, either:

- Continue to pass the `--print-graph` flag alongside your new flag.
- Add support in `snyk-maven-plugin` and `snyk-gradle-plugin` for your new flag.

snyk-abedonik: Fixed

mcombuechen (approved):

PR #145, opened 2026-04-16
feat: (cmpa-502) bun resolver implementation

cjheppell:
Adds a new `bun` plugin that resolves dependency graphs for Bun projects without bespoke lockfile parsing.

### Approach

Rather than parsing `bun.lock` directly, the plugin shells out to `bun why '*' --top` (introduced in bun v1.2.19), which outputs the full reverse-dependency graph for every installed package in one pass. The output is parsed into a `reverseGraph` (a map of `"name@version"` → set of dependents), inverted into a forward adjacency map, and DFS-walked to produce Snyk dep graphs.

We delegate to `bun why` to avoid having to parse the lockfile and infer node modules resolution behaviour.

It also ensures we align with other bun behaviour. For example - Bun installs peer dependencies by default, so we do not omit peer dependencies from the dep graph. If it shows up in bun why, it shows up in the dep graph.

### Workspace support

Workspace projects produce one `SCAResult` per workspace package plus one for the root, matching how the npm/yarn plugins handle monorepos. Each workspace package is treated as a leaf in every graph it appears in
as a dependency; its transitive subtree is only walked in its own dedicated graph. This prevents duplicate vulnerability reporting across workspace packages.

In the future, we also want to introduce a new component to represent the top-level bun.lock which will capture the full workspaces dependency graph as a unit.

### Unit test fixtures

`testdata/simple/` and `testdata/workspace/` provide synthetic `why_output.txt` files used by `plugin_test.go` and `why_parser_test.go`. Full acceptance tests with real `bun.lock` golden files live in the
companion PR (see below).

### Known limitation

Packages bundled inside another (via `bundledDependencies`) can be subject to an active [bun bug (#29263)](https://redirect.github.com/oven-sh/bun/issues/29263). There is a `TestBundledDepsVisibleInBunWhy` test that will self-activate and stop skipping when bun ships a fix. If it would be preferred for this test to _fail_ so that we know when this has been fixed, then we can also do that.

---

> **Companion PR:** The acceptance test suite with golden fixture files (~142k lines of `bun.lock` and `expected.json`) is in a separate PR branched off this one to keep this diff reviewable.

### DIFF
--- a/pkg/ecosystems/javascript/bun/package_json.go
+++ b/pkg/ecosystems/javascript/bun/package_json.go
@@ -1,0 +1,45 @@
+package bun
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const (
+	packageJSONFile = "package.json"
+	bunLockFile     = "bun.lock"
+	defaultVersion  = "0.0.0"
+)
+
+// packageJSON represents the fields we need from a package.json file.
+type packageJSON struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
+
+// readPackageJSON reads and parses the package.json in the given directory.
+func readPackageJSON(dir string) (*packageJSON, error) {
+	path := filepath.Join(dir, packageJSONFile)
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading %s: %w", path, err)
+	}
+
+	var p packageJSON
+	if err := json.Unmarshal(data, &p); err != nil {
+		return nil, fmt.Errorf("parsing %s: %w", path, err)
+	}
+
+	if p.Name == "" {
+		return nil, fmt.Errorf("%s: missing required 'name' field", path)
+	}
+
+	if p.Version == "" {
+		p.Version = defaultVersion
+	}
+
+	return &p, nil
+}
--- a/pkg/ecosystems/javascript/bun/package_json.go
+++ b/pkg/ecosystems/javascript/bun/package_json.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+
+	snykecosystems "github.com/snyk/error-catalog-golang-public/opensource/ecosystems"
+	"github.com/snyk/error-catalog-golang-public/snyk_errors"
 )
 
 const (
@@ -25,16 +28,21 @@ func readPackageJSON(dir string) (*packageJSON, error) {
 
 	data, err := os.ReadFile(path)
 	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, snykecosystems.NewUnprocessableFileError(
+				fmt.Sprintf("%s not found: %v", packageJSONFile, err),
+				snyk_errors.WithCause(err),
+			)
+		}
 		return nil, fmt.Errorf("reading %s: %w", path, err)
 	}
 
 	var p packageJSON
 	if err := json.Unmarshal(data, &p); err != nil {
-		return nil, fmt.Errorf("parsing %s: %w", path, err)
-	}
-
-	if p.Name == "" {
-		return nil, fmt.Errorf("%s: missing required 'name' field", path)
+		return nil, snykecosystems.NewUnparseableManifestError(
+			fmt.Sprintf("Failed to parse %s: %v", packageJSONFile, err),
+			snyk_errors.WithCause(err),
+		)
 	}
 
 	if p.Version == "" {
--- a/pkg/ecosystems/javascript/bun/plugin.go
+++ b/pkg/ecosystems/javascript/bun/plugin.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 
 	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems"
 	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems/discovery"
 	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems/logger"
+	"github.com/snyk/cli-extension-dep-graph/pkg/identity"
 )
 
 const logFieldLockFile = "lockFile"
@@ -83,10 +83,19 @@ func (p Plugin) buildResults(
 	lockFileRelPath, lockFileAbsDir string,
 	exec bunWhyRunner,
 ) []ecosystems.SCAResult {
-	meta := ecosystems.Metadata{TargetFile: lockFileRelPath}
-
+	// TargetFile for error results: the root package.json alongside bun.lock.
+	lockFileDir := filepath.Dir(lockFileRelPath)
+	rootTargetFile := filepath.Join(lockFileDir, packageJSONFile)
 	errResult := func(err error) []ecosystems.SCAResult {
-		return []ecosystems.SCAResult{{Metadata: meta, Error: err}}
+		return []ecosystems.SCAResult{{
+			ProjectDescriptor: identity.ProjectDescriptor{
+				Identity: identity.ProjectIdentity{
+					Type:       pkgManager,
+					TargetFile: &rootTargetFile,
+				},
+			},
+			Error: err,
+		}}
 	}
 
 	log.Info(ctx, "Building bun dependency graph", logger.Attr(logFieldLockFile, lockFileRelPath))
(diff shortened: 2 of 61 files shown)


pkg/ecosystems/javascript/bun/plugin.go
```diff
@@ -0,0 +1,230 @@
...
+	defer output.Close()
+
+	out, err := parseWhyOutput(ctx, log, output)
+	if err != nil {
+		return errResult(fmt.Errorf("parsing bun why output: %w", err))
+	}
+
+	log.Debug(ctx, "Parsed bun why output", logger.Attr(logFieldLockFile, lockFileRelPath), logger.Attr("packages", len(out.Graph)))
+
+	graphResults, err := buildDepGraphs(pkgJSON.Name, pkgJSON.Version, out)
```
james-snyk, line 131: Might be worth handling the fallback root package name like the other ecosystem do (when name is not set we use package.json)

pkg/ecosystems/javascript/bun/plugin.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems/logger"
+	"github.com/snyk/cli-extension-dep-graph/pkg/identity"
+)
+
+const logFieldLockFile = "lockFile"
+
+// Plugin implements ecosystems.SCAPlugin for Bun projects.
+// It uses `bun why '*' --top` to resolve the full dependency graph without
+// bespoke lockfile parsing, requiring bun >= 1.2.19.
+type Plugin struct {
```
james-snyk, line 21: Should we rename to Resolver now instead of later?

pkg/ecosystems/javascript/bun/package_json.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("reading %s: %w", path, err)
+	}
+
+	var p packageJSON
+	if err := json.Unmarshal(data, &p); err != nil {
+		return nil, fmt.Errorf("parsing %s: %w", path, err)
+	}
+
+	if p.Name == "" {
```
james-snyk: Do we want to make this required? Other plugins default to the file name instead.

pkg/ecosystems/javascript/bun/package_json.go
```diff
@@ -0,0 +1,45 @@
...
+	path := filepath.Join(dir, packageJSONFile)
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading %s: %w", path, err)
+	}
+
+	var p packageJSON
+	if err := json.Unmarshal(data, &p); err != nil {
+		return nil, fmt.Errorf("parsing %s: %w", path, err)
```
james-snyk: could we return the NewUnprocessableFileError/NewSyntaxIssuesError snyk error instead here

pkg/ecosystems/javascript/bun/plugin.go
```diff
@@ -0,0 +1,230 @@
...
+		for _, r := range fileResults {
+			if r.Error != nil {
+				log.Error(ctx, "Failed to build bun dependency graph",
+					logger.Attr(logFieldLockFile, file.RelPath),
+					logger.Err(r.Error),
+				)
+			}
+		}
+		results = append(results, fileResults...)
+		processedFiles = append(processedFiles, file.RelPath)
```
james-snyk, line 71: Currently we only append the lock file but we probably also need to include the package.json file else other plugins will pick them up and scan them again

pkg/ecosystems/javascript/bun/package_json.go
```diff
@@ -0,0 +1,45 @@
...
+	Version string `json:"version"`
+}
+
+// readPackageJSON reads and parses the package.json in the given directory.
+func readPackageJSON(dir string) (*packageJSON, error) {
+	path := filepath.Join(dir, packageJSONFile)
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading %s: %w", path, err)
```
james-snyk, line 37: if the file is not found should we return NewManifestNotFoundError?

pkg/ecosystems/javascript/bun/package_json.go
calhar-snyk: Indeed. Falling back to the directory name is probably reasonable.

pkg/ecosystems/javascript/bun/plugin.go
```diff
@@ -0,0 +1,230 @@
...
+	dir string,
+	options *ecosystems.SCAPluginOptions,
+) ([]discovery.FindResult, error) {
+	if options == nil {
+		options = ecosystems.NewPluginOptions()
+	}
+
+	switch {
+	case options.Global.TargetFile != nil:
+		if filepath.Base(*options.Global.TargetFile) != bunLockFile {
```
calhar-snyk, line 186: Tangentially related to @james-snyk question above re: reporting `package.json` as a processed file.

Would it make sense for us to tolerate `snyk test --file=foo/package.json` and expect it to be correctly handled as bun?

pkg/ecosystems/javascript/bun/plugin.go
```diff
@@ -0,0 +1,230 @@
...
+			return nil, nil
+		}
+
+		files, err := discovery.FindFiles(ctx, dir, discovery.WithTargetFile(*options.Global.TargetFile))
```
calhar-snyk, line 190: Nitpick -- not specifically for the changes here but for the file discovery in general.

The current CLI uses a default depth for discovery of (I think) 4. It doesn't currently look like we're currently supporting the `--detection-depth` flag? Should we?

pkg/ecosystems/javascript/bun/why_parser.go
```diff
@@ -0,0 +1,240 @@
...
+	//
+	// Group 1: "name@version" of the dependent package.
+	//   Root project references (e.g. "my-app (requires ^4)") have no version component,
+	//   so they never match — they are caught by depth1RootRe instead.
+	depth1LineRe = regexp.MustCompile(`^  [└├]─ (?:dev |peer |optional )?(@?[^@\s(]+@\S+?)(?:\s+\(requires[^)]+\))?$`)
+
+	// depth1RootRe matches a depth-1 line where the dependent is the root project.
+	// The dependent has no @version component — bun prints just the project name.
+	//
+	// Group 1: "dev " if this is a dev dependency; empty string otherwise.
```
calhar-snyk, line 42: ~I don't think this is accurate?~
```
(dev )?(?:peer |optional )?
```
~Compared with the following from `depth1LineRe`~
```
(?:dev |peer |optional )?
```

Nevermind, I'm an idiot and forgot that `(?:)` is a non-capturing group

calhar-snyk: Followup though. Do we care about optional dependencies and being able to exclude/include them?

pkg/ecosystems/javascript/bun/plugin.go
cjheppell: I'd rather rename this wholesale, because it would also involve a breaking API change if we wanted to rename the interface (`SCAPlugin` -> `SCAResolver`).

Leaving things inconsistent in this PR feels like the worst of both worlds.

pkg/ecosystems/javascript/bun/plugin.go
cjheppell: Good spot.

pkg/ecosystems/javascript/bun/why_parser.go
cjheppell: By default, bun _does_ install optional dependencies. So I think for consistency with that behaviour we should include them: https://bun.com/docs/pm/cli/install#omitting-dependencies

JamesPatrickGill: I think if not installed they are not involved in the lock at all and "why" doesn't even show them. But I could be wrong

cjheppell: Seems like `bun why '*'` always prints out the optional/peer even if you run with `--omit`. They also appear in the lockfile. Despite that, the node modules _doesn't_ include the optional/peer deps. 

So... somewhat peculiar behaviour. But we'll have to respect whatever the lockfile/output of bun why states.

pkg/ecosystems/javascript/bun/plugin.go
cjheppell: So, this is interesting.
When the root package doesn't contain a name, bun why produces an output that makes root level direct dependencies invisible to the parser. As such, for the time being we're enforcing the requirement of a name on the package.json.

This will ensure that we produce accurate dep graphs, and fail loudly if we can't do so. There might be something we can do in the future if this proves to be problematic.

pkg/ecosystems/javascript/bun/package_json.go
cjheppell: see https://github.com/snyk/cli-extension-dep-graph/pull/145#discussion_r3124029334

pkg/ecosystems/javascript/bun/plugin.go
cjheppell: I don't think so.

The identification of a project being "bun" is the presence of a bun.lock. That's the only reasonable heuristic we can use.

We should be reporting the package.json's as having being processed by this bun resolver, to avoid duplication from other resolvers. But targeting package.json on its own doesn't feel appropriate for bun imo.

pkg/ecosystems/javascript/bun/plugin.go
cjheppell: It's a good point. Currently, the detection depth appears to only makes it way to the discovery of manifest files in this plugin. Instead, it gets passed to the "legacy" flow.

We should handle this, but its a concern beyond the scope of this PR for now.

pkg/ecosystems/javascript/bun/why_parser.go
JamesPatrickGill: Thats an interesting choice

JamesPatrickGill (approved):

PR #146, opened 2026-04-16
chore: (cmpa-502) bun resolver fixture tests

cjheppell:
> [!IMPORTANT]
> **Stacked on:** `cmpa-502/bun-resolver-implementation`. Review that PR first.

### What's in this PR

Two additions on top of the implementation PR:

**Acceptance tests** (`acceptance_test.go`, `stability_test.go`)

- `TestAcceptance` — runs the full plugin against every non-workspace fixture and compares output against a golden `expected.json`
- `TestAcceptance_Workspaces` — same, but for workspace projects; compares one golden file per workspace package plus the root
- `TestInvalidPackageJSON` — verifies a directory with a malformed `package.json` produces no results rather than a crash
- `TestBundledDepsVisibleInBunWhy` — known-failing test for a [bun bug (#29263)](https://redirect.github.com/oven-sh/bun/issues/29263) where packages bundled inside another package's dist are invisible to `bun why`; skips itself until bun ships a fix, at which point it will self-activate
- `TestBunWhyOutputStability` — see below

**Daily stability check** (`stability_test.go`, `.github/workflows/bun-why-stability.yml`)

Runs on a schedule against the **latest bun release** (not a pinned version). For every fixture directory it asserts that the plugin produces no errors and that every package in `bun.lock` appears in at least one
dep graph. This catches breaking changes to `bun why '*' --top` output format before they reach production users. Unlike `TestAcceptance`, it does not compare against golden files, so it won't false-positive on
minor version bumps in resolved packages.

### Fixtures

Golden files cover the following scenarios across ~20 fixture directories:

| Fixture | What it exercises |
|---|---|
| `one-dep`, `simple-*` | Basic resolution |
| `goof` | Real-world package with many transitive deps |
| `only-dev-deps`, `empty-dev-deps` | Dev dep handling |
| `different-versions` | Multiple versions of the same package |
| `cyclic-dep` | Cycle protection |
| `deeply-nested-packages`, `deeply-scoped` | Deep trees, scoped packages |
| `local-pkg`, `local-pkg-without-workspaces` | `file:` protocol dependencies |
| `override-with-*`, `simple-override`, `simple-version-range-override`, `nested-non-alias-*` | npm overrides |
| `dist-tag-sub-dependency`, `npm-scoped-sub-dep` | Dist tags, scoped sub-deps |
| `missing-*` | Optional/missing deps |
| `workspaces`, `workspaces-glob-*`, `workspaces-packages` | Workspace glob patterns |
| `workspaces-cross-dep` | Cross-workspace dependencies |
| `workspaces-catalog` | Bun `catalog:` / `catalog:<name>` protocol |
| `bundled-deps-wasm`, `bundled-top-level-dep` | Bundled deps (currently skipped, tracking bun #29263) |

All `bun.lock` files were generated with `bun install --save-text-lockfile` against bun ≥ 1.2.19.

### Updating golden files

If the plugin output changes intentionally:

```bash
go test -run 'TestAcceptance|TestAcceptance_Workspaces' ./pkg/ecosystems/javascript/bun/... -update

### CODE REVIEW

File: pkg/ecosystems/javascript/bun/stability_test.go [hunk 1]
```diff
@@ -0,0 +1,89 @@
...
+//
+//  1. BuildDepGraphsFromDir returns no error and no per-result errors.
+//  2. Every package listed in bun.lock appears in at least one dep graph.
+//
+// Condition 2 catches silent regressions where the parser produces a valid but
+// incomplete graph — e.g. if bun changes indentation or tree characters so that
+// depth-1 lines no longer match the regex.
+func TestBunWhyOutputStability(t *testing.T) {
+	if _, err := exec.LookPath("bun"); err != nil {
+		t.Skip("bun not found in PATH — skipping stability test")
```
calhar-snyk: This doesn't feel like it should be a skip

pkg/ecosystems/javascript/bun/acceptance_test.go
```diff
@@ -0,0 +1,406 @@
...
+var updateGolden = flag.Bool("update", false, "overwrite expected.json golden files with current plugin output")
+
+// runAcceptanceSuite discovers fixtures under fixtureDir and asserts each one
+// produces a dep graph matching its expected.json. Only the first result is
+// compared — workspace sub-graphs are covered by TestAcceptance_Workspaces.
+func runAcceptanceSuite(t *testing.T, fixtureDir string, opts *ecosystems.SCAPluginOptions) {
+	t.Helper()
+
+	if _, err := exec.LookPath("bun"); err != nil {
+		t.Skip("bun not found in PATH — skipping acceptance suite")
```
calhar-snyk: Ditto. Doesn't feel like it should be a skip

pkg/ecosystems/javascript/bun/acceptance_test.go
```diff
@@ -0,0 +1,406 @@
...
+		rootPkg string // root package name identifies which result to compare
+		golden  string // filename of the golden file inside the fixture dir
+	}
+
+	tests := []struct {
+		fixture string
+		graphs  []graphCase
+	}{
+		{
+			fixture: "workspaces",
```
calhar-snyk, line 136: These all seem fairly standardised and discoverable like we're doing with the other tests. The only ones that might cause issue are the `my-monorepo` and catalog monorepo ones?

pkg/ecosystems/javascript/bun/stability_test.go
cjheppell: It's a good point... I think I was optimising for "as this repository grows, there might be an ever increasing set of dependent binaries, and should we really be making people fail if they don't have them?".

In other words, if you come along to this repo, `go test ./...` will skip any of the tests that you don't have the requisite binaries for (uv/bun/maven possibly in the future, etc)..

But in hindsight, probably better to fail loudly.

calhar-snyk (approved):

PR #149, opened 2026-04-21
feat: emit root package name on project identity

mcombuechen:
### What this does

Since the os-flows extension needs the root package name for identity, let's add it to this extensions interface so that os-flows don't have to extract it from the dep-graph.

Opinionated change: also renames `Type` to `ProjectType`.

### DIFF
--- a/pkg/ecosystems/orchestrator/orchestrator.go
+++ b/pkg/ecosystems/orchestrator/orchestrator.go
@@ -212,7 +212,8 @@ func depGraphOutputToSCAResult(output *parsers.DepGraphOutput, ictx workflow.Inv
 	result := ecosystems.SCAResult{
 		ProjectDescriptor: identity.ProjectDescriptor{
 			Identity: identity.ProjectIdentity{
-				TargetFile: &output.NormalisedTargetFile,
+				TargetFile:    &output.NormalisedTargetFile,
+				TargetRuntime: output.TargetRuntime,
 			},
 		},
 	}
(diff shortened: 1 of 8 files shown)


pkg/ecosystems/orchestrator/orchestrator.go
```diff
@@ -212,7 +212,8 @@ func depGraphOutputToSCAResult(output *parsers.DepGraphOutput, ictx workflow.Inv
 	result := ecosystems.SCAResult{
 		ProjectDescriptor: identity.ProjectDescriptor{
 			Identity: identity.ProjectIdentity{
-				TargetFile: &output.NormalisedTargetFile,
+				TargetFile:    &output.NormalisedTargetFile,
+				TargetRuntime: output.TargetRuntime,
```
james-snyk, line 216: We probably dont want to set these values here as it should be based on which ecosystem it

mcombuechen: So the `TargetFile` change isn't new. Do you think we should move it into the `if dg.PkgManager.Name != ""` branch?
About the `TargetRuntime`, that one is optional so I think it will only ever be populated when the ecosystem sets it, so this should be fine.
Let me know how you think we should swing this.

mcombuechen: Furthermore, I don't understand why we even have the `if dg.PkgManager.Name != ""` check. I don't think we'd ever process dep-graphs without a package manager name.

james-snyk (approved):

PR #150, opened 2026-04-21
feat: add gradle plugin [CMPA-473]

tonidevine1:

### Summary

Adds a native Gradle dependency graph resolver under `pkg/ecosystems/gradle/`. The plugin injects an embedded init script into the user's Gradle build, runs a custom `snykDependencyGraph` task that walks Gradle's `ResolutionResult` API, and converts the resulting JSON into a Snyk dep-graph.

The plugin is wired in via `SCAPluginOptions` (new `GradleOptions` struct: `--configuration-matching`, `--gradle-sub-project`, `--all-sub-projects`, `--init-script`, `--gradle-skip-wrapper`) but this PR does not yet hook it into the orchestrator — that lands separately.

#### Gradle initscript

The heart of the resolver is `pkg/ecosystems/gradle/init/snyk-deps-init.gradle`, embedded into the Go binary via `//go:embed` and written to a temp file at invocation time. It registers a `snykDependencyGraph` task on the root project that, for each project and each resolvable configuration, walks `configurations.incoming.resolutionResult` and emits a per-project JSON entry containing the dependency tree plus a flat list of all resolved components.

The script targets Gradle 6 - 9 with compatibility shims for API differences across that range; Gradle 5 and earlier are explicitly unsupported. Output is streamed to disk rather than accumulated in memory to keep large multi-module builds bounded, and the path of the generated file is communicated back to the Go side via a `SNYK_DEPS_JSON` stdout marker.

The Go side invokes Gradle with `--no-daemon --no-parallel --console=plain -Dorg.gradle.welcome=never :snykDependencyGraph`. `--no-parallel` is unfortunate for performance and is flagged in the README as something to revisit.

#### Project discovery and deduplication

Three discovery modes, based on options:

1. **`--target-file`**: validate it's a Gradle build file, then point discovery at it.
2. **`--all-projects`**: recursively find all `build.gradle{,.kts}` and `settings.gradle{,.kts}`, applying common excludes plus user-supplied ones.
3. **Default**: the highest-priority Gradle file in the root only (`build.gradle` > `build.gradle.kts` > `settings.gradle` > `settings.gradle.kts`).

The trickier part is that one Gradle invocation against a root project naturally produces results for all sub-projects. Naively iterating discovered files would re-invoke Gradle once per sub-module, which is both slow and produces duplicate results.

The fix is directory-based deduplication keyed off the relative project path: files are sorted by path depth so parent directories run first, and after each Gradle invocation every directory whose build file appears in that run's output is recorded so subsequent discovered files in those directories are skipped.

Settings files get special handling: they're treated as project-root markers (Gradle is run from there to discover sub-projects) but aren't themselves considered scan targets, so they're not passed through as the build-file fallback during result conversion.

For `--target-file` specifically, we still run the task on the root project to leverage Gradle's native multi-module handling, then *filter* the resulting projects by absolute build-file path so only the requested project's graph is returned. This avoids reimplementing Gradle's project resolution in Go.

#### Gradle binary resolution

`ResolveGradleBinary` walks up the directory tree from the project dir looking for `gradlew` (or `gradlew.bat` on Windows) and only falls back to `gradle` on `PATH` if no wrapper is found or `--gradle-skip-wrapper` is set. Walking up matters because in multi-module layouts the wrapper typically sits at the repo root, not in each sub-project.

#### Configuration handling

The resolver is deliberately non-prescriptivist about which Gradle configurations to include. All resolvable configurations across all projects are merged into a single dep-graph per project. The trade-off is documented in the README, but in short:

- Plugins introduce arbitrary configurations with arbitrary semantics, so hardcoding a "runtime-only" filter would either miss legitimate dependencies or whitelist plugin-specific ones we don't know about.
- Build-time dependencies (annotation processors, code generators, test frameworks) are part of the supply chain and worth surfacing.
- The downside is that the same transitive may appear at multiple versions when configurations resolve differently (e.g. `compileClasspath` pinning `guava:30` while `testRuntimeClasspath` resolves `guava:31`). We use a fresh `seen` map per configuration when merging, so different versions from different configurations both make it into the final graph rather than the first one winning.

`--configuration-matching` is parsed and reserved for future use; configuration-scoped filtering will land when we tackle full feature parity.

#### Graph construction

Edge insertion is deduplicated via a `(parent → set of children)` map wrapping `depgraph.Builder.ConnectNodes`, since that builder isn't idempotent and a dep present in N configurations would otherwise produce N parallel edges from the same parent.

#### Init script augmentation

User-supplied `--init-script` is treated as augmentation, not replacement: the embedded Snyk init script is always passed via `--init-script`, and any user-supplied init script is appended as an additional `--init-script` flag. Replacing the embedded script would break dependency extraction entirely since the JSON output depends on it.

#### Classifiers

Currently dropped. Dependency IDs are `group:artifact:version` with no classifier component, so `commons-io:commons-io:2.6:sources` and `commons-io:commons-io:2.6` collapse to the same node. Documented in the README as a known limitation; upstream services don't currently consume classifier info.

#### Error surfacing

Gradle binary resolution failures and Gradle execution failures don't fail the whole scan: each is wrapped into an `SCAResult` with `Error` set and a relative `TargetFile`, so the caller can report it as a per-project warning rather than a hard failure. JSON parse errors of the init script's output *are* fatal, since they indicate a bug in the script rather than a user build problem.


pkg/ecosystems/gradle/depgraph.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/snyk/dep-graph/go/pkg/depgraph"
+)
+
+const pkgManagerName = "gradle"
+
+// preferredConfigs lists Gradle configuration names tried in order when no
+// explicit --configuration-matching filter is provided.  runtimeClasspath
+// gives the most complete picture of what actually runs; compileClasspath is
+// a common fallback.
+var preferredConfigs = []string{
```
calhar-snyk: All the configuration handling pieces here should be left for later when we're adding feature parity. The basic implementation should basically just merge all the configurations.

pkg/ecosystems/orchestrator/orchestrator.go
```diff
@@ -34,19 +35,28 @@ func ResolveDepgraphs(ictx workflow.InvocationContext, dir string, opts ecosyste
 	enhancedLogger := ictx.GetEnhancedLogger()
 
 	pythonResult := resolvePython(ictx.Context(), enhancedLogger, dir, opts)
+	gradleResult := resolveGradle(ictx.Context(), enhancedLogger, dir, opts)
```
calhar-snyk: In hindsight, we probably don't want to be doing this yet.

pkg/ecosystems/gradle/init/snyk-deps-init.gradle
```diff
@@ -0,0 +1,237 @@
...
+                outputDir = root.layout.buildDirectory.dir("reports").get().asFile
+            } catch (Exception e) {
+                outputDir = file("${root.buildDir}/reports")
+            }
+            outputDir.mkdirs()
+            def outputFile = file("${outputDir}/snyk-dependency-graph.json")
+
+            // Stream JSON output to avoid memory accumulation for large projects
+            outputFile.withWriter('UTF-8') { writer ->
+                writer.println('{')
```
cjheppell: I notice we're attempting to "pretty print" this to some degree. Do we need to?
`println` could be `print` and whitespace could be removed, right?

pkg/ecosystems/gradle/init/snyk-deps-init.gradle
```diff
@@ -0,0 +1,237 @@
...
+            try {
+                outputDir = root.layout.buildDirectory.dir("reports").get().asFile
+            } catch (Exception e) {
+                outputDir = file("${root.buildDir}/reports")
+            }
+            outputDir.mkdirs()
+            def outputFile = file("${outputDir}/snyk-dependency-graph.json")
+
+            // Stream JSON output to avoid memory accumulation for large projects
+            outputFile.withWriter('UTF-8') { writer ->
```
cjheppell, line 41: what about streaming out to an ndjson file?
that avoids us having to handle open/closing brackets and "first" tracking.

pkg/ecosystems/gradle/init/snyk-deps-init.gradle
```diff
@@ -0,0 +1,237 @@
...
+        maxMemoryMB     : runtime.maxMemory() / (1024 * 1024) as long,
+    ]
+}
+
+// ── Per-project entry ───────────────────────────────────────────────
+
+def buildProjectEntry(project) {
+    def resolvableConfigs = project.configurations.matching { it.canBeResolved }
+
+    def gav = "${project.group}:${project.name}:${project.version}"
```
cjheppell, line 90: For my understanding - what does `gav` stand for here?
I thought it was an acronym for `group, name, version` but that would be `gnv` 🤔

pkg/ecosystems/gradle/binary_test.go
```diff
@@ -0,0 +1,159 @@
...
+		require.NoError(t, os.WriteFile(gradlewBatPath, []byte("@echo off"), 0o755))
+
+		binary, err := ResolveGradleBinary(dir, false)
+		require.NoError(t, err)
+
+		if runtime.GOOS == "windows" {
+			assert.Equal(t, gradlewBatPath, binary)
+		} else {
+			assert.Equal(t, gradlewPath, binary)
+		}
```
cjheppell: it would be nicer to "mock" this value (`runtime.GOOS`) so that we can exercise both paths and not rely on actually executing this on a windows/unix environment to test.

pkg/ecosystems/gradle/binary_test.go
```diff
@@ -0,0 +1,159 @@
...
+	})
+
+	t.Run("finds wrapper in parent directory", func(t *testing.T) {
+		parentDir := t.TempDir()
+		childDir := filepath.Join(parentDir, "subproject")
+		require.NoError(t, os.MkdirAll(childDir, 0o755))
+
+		// Create wrapper in parent directory
+		var wrapperName string
+		if runtime.GOOS == "windows" {
```
cjheppell: dittto

pkg/ecosystems/gradle/depgraph_test.go
```diff
@@ -0,0 +1,398 @@
...
+		assert.True(t, ids["com.google.guava:guava@32.1.2-jre"], "deps from valid config should be present")
+	})
+
+	t.Run("returns empty graph when all configurations have errors", func(t *testing.T) {
+		proj := makeProject("com.example", "app", "1.0.0", []gradleConfig{
+			{Name: "runtimeClasspath", Error: "resolution failed"},
+		})
+
+		dg, err := buildDepGraph(&proj)
+		require.NoError(t, err)
```
cjheppell, line 180: Curious on semantics -- if all configs have errors, we currently do not return an error here, but an empty graph? I suppose "building the dep graph" didn't fail, but the config we tried to build it from did? 🤔

pkg/ecosystems/gradle/executor.go
```diff
@@ -0,0 +1,83 @@
...
+	}, extraArgs...)
+
+	cmd := exec.CommandContext(ctx, gradleBinary, args...)
+	cmd.Dir = projectDir
+
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
```
cjheppell: presumably because stdout should only really be the marker, this is fine to read the entire contents into memory and not require an `io.Reader` ?

pkg/ecosystems/gradle/executor.go
```diff
@@ -0,0 +1,83 @@
...
+
+	data, err := os.ReadFile(outputFile)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"gradle ran successfully but could not read output file %s: %w\nstdout:\n%s",
+			outputFile, err, stdout.String(),
+		)
+	}
+
+	return data, nil
```
cjheppell: I suspect we ought to be returning an `io.Reader` (or `io.ReadCloser`) here rather than the entire byte array.

pkg/ecosystems/gradle/plugin.go
```diff
@@ -0,0 +1,458 @@
...
+	files []discovery.FindResult,
+	dir string,
+	options *ecosystems.SCAPluginOptions,
+) ([]ecosystems.SCAResult, []string, error) {
+	// Sort by path depth to process parent directories first
+	sort.Slice(files, func(i, j int) bool {
+		depthI := strings.Count(files[i].Path, string(filepath.Separator))
+		depthJ := strings.Count(files[j].Path, string(filepath.Separator))
+		return depthI < depthJ
+	})
```
cjheppell: be aware that this is modifying the slice passed in, which will be visible to the caller.
fine if intended, but might not be

cjheppell (commented):
Some comments, but no major blockers.

pkg/ecosystems/gradle/init/snyk-deps-init.gradle
calhar-snyk: 100% yes. This is mostly a holdover from spiking.

pkg/ecosystems/gradle/init/snyk-deps-init.gradle
calhar-snyk: Indeed. Will unprettify in this. And do ndjson in the script improving PR

pkg/ecosystems/gradle/init/snyk-deps-init.gradle
calhar-snyk: Close. `group`, `artifact`, `version`.

pkg/ecosystems/gradle/executor.go
calhar-snyk: Theoretically, yes. Practically, we should stream this and I've just left it by the wayside. Clients can do completely arbitrary things in gradle, so they could have a buildscript that insists on writing a billion 'A's before any other output.

cjheppell (approved):

PR #151, opened 2026-04-22
feat: include root component name in project identity for pip, pipenv…

paulrosca-snyk:
…, and uv plugins

### What this does

Includes root component name in all plugins and gets rid of `baseNameOverride`

### DIFF
--- a/pkg/ecosystems/python/pip/plugin.go
+++ b/pkg/ecosystems/python/pip/plugin.go
@@ -67,7 +67,7 @@ func (p Plugin) BuildDepGraphsFromDir(
 	for _, file := range files {
 		g.Go(func() error {
 			projectName := GetProjectName(file.RelPath, dir)
-			result, err := p.buildDepGraphFromFile(ctx, log, file, pythonVersion, options.Python.NoBuildIsolation, projectName, options.Global.ProjectName)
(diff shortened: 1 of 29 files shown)


mcombuechen (approved):

pkg/ecosystems/python/pip/plugin.go
james-snyk, line 194: This was missed but for pip/pipenv we need to pass the baseNameOverride to the depgraph as we set the root node to be the value of the override for python.

mcombuechen: We examined the code and the `baseNameOverride` wasn't used anywhere; it was merely the value of `--project-name` and echoed back from the plugin. The value going into the graph's root package was that of `projectName`.
Are you saying that this was not quite correct and that the value of `--project-name` should go into the root package?

james-snyk: Just for the python ecosystem the root node gets set to the value of --project-name

mcombuechen: Which should still work, there's the argument `projectName` which goes into the dep-graph. That one is set to the value of `--project-name`.

james-snyk: Sorry thats what the issue is as currently project name var is just based on directory name. GetProjectName needs to also take the flag value. Just noticed that this was forgotten when reviewing this PR.

james-snyk (approved):

PR #152, opened 2026-04-22
feat: use --exclude-paths for processed file exclusions

danielroymoore:

## What this does

Introduces support for a new `--exclude-relative` CLI flag that accepts **relative paths** (comma-separated) for excluding files and directories during multi-project scans. This replaces the lossy `path.Split` workaround from PR #147, which stripped paths down to basenames before forwarding them to the legacy CLI's `--exclude` flag.

### The problem

The existing `--exclude` flag only accepts basenames (e.g. `package.json`, `node_modules`). When the orchestrator hands off "processed files" to the legacy CLI for fallback resolution, it needs to tell the CLI *which specific files* to skip. PR #147 worked around this by calling `path.Split` to extract just the filename:

```
processedFiles before path.Split:  packages/api/package.json, packages/web/package.json
processedFiles after path.Split:   package.json, package.json
```

This caused **all** `package.json` files to be excluded, breaking workspace-style projects (bun, uv, yarn workspaces) where multiple packages share the same manifest filename but live at different paths.

The UV plugin had an additional workaround that cleared `ProcessedFiles` entirely to avoid this bug, which meant processed files were never properly excluded from fallback resolution.

### What changed

**Flag registration & forwarding**
- Registered `--exclude-relative` in `flags.go` as a string flag
- `legacy_resolution.go` forwards it to the legacy CLI when set
- `sbom_resolution.go` parses it via `parseExcludeFlag` and passes it to plugin options via `WithExcludeRelative()`

**Orchestrator (`LegacyFallback`)**
- Removed the `path.Split` loop — processed files now retain their relative paths
- Processed files are forwarded using `--exclude-relative=` instead of `--exclude=`
- Any user-provided `--exclude-relative` from `RawFlags` is **merged** with the auto-generated processed file exclusions, preventing user flags from being silently dropped

**SBOM resolution (`applyProcessedFilesExclusions`)**
- Now writes to `FlagExcludeRelative` instead of `FlagExclude`
- Merges with any existing `--exclude-relative` value rather than clobbering it
- Simplified string building using `strings.Join` on a pre-allocated slice

**Plugin options**
- Added `ExcludeRelative CommaSeparatedString` field to `GlobalOptions`
- Added `WithExcludeRelative()` builder method on `SCAPluginOptions`

**UV plugin cleanup**
- Removed the `if true { result.ProcessedFiles = []string{} }` workaround and its TODO comment — the UV plugin now correctly reports processed files

### Companion PR

This depends on [snyk/cli#6741](https://github.com/snyk/cli/pull/6741), which implements `--exclude-relative` in the legacy CLI's `find-files.ts` with proper validation (rejects absolute paths and `..` references) and path-based exclusion in the file discovery engine.

## Trade-offs & callouts

1. **`--exclude` behaviour is unchanged.** This is a purely additive change. The existing `--exclude` flag continues to work as before (basename-only matching). No existing CLI contracts are modified.

2. **Go plugins don't yet consume `--exclude-relative` internally.** The Go-side file discovery (`files.go`) already supports relative path glob matching via `filepath.Match`, but individual plugins don't filter against `ExcludeRelative` during their own discovery phase. This means the flag currently only takes effect during legacy CLI fallback and SBOM orchestration. Adding per-plugin filtering is a natural follow-up but is not a regression — the current behaviour is equivalent to what existed before this PR.

3. **Flag merging in `LegacyFallback`.** When a user explicitly passes `--exclude-relative=foo` and the orchestrator also has processed files to exclude, both are merged into a single `--exclude-relative` argument. This is intentional to avoid silently shadowing user input. The merge is simple string concatenation of comma-separated values.

4. **No `ARG_MAX` risk in practice.** Processed file paths are relative and typically short. Even in very large monorepos, the combined `--exclude-relative` argument is unlikely to approach OS argument length limits.

### DIFF
--- a/internal/legacycli/legacy_resolution.go
+++ b/internal/legacycli/legacy_resolution.go
@@ -115,6 +115,11 @@ func PrepareLegacyFlags(argument string, cfg configuration.Configuration, logger
 		logger.Print("Exclude:", exclude)
 	}
 
+	if excludePaths := cfg.GetString(workflow.FlagExcludePaths); excludePaths != "" {
+		cmdArgs = append(cmdArgs, "--exclude-paths="+excludePaths)
+		logger.Print("Exclude paths:", excludePaths)
+	}
+
 	if detectionDepth := cfg.GetString(workflow.FlagDetectionDepth); detectionDepth != "" {
 		cmdArgs = append(cmdArgs, "--detection-depth="+detectionDepth)
 		logger.Print("Detection depth:", detectionDepth)
--- a/internal/workflow/flags.go
+++ b/internal/workflow/flags.go
@@ -6,6 +6,7 @@ const (
 	FlagPrintOutputJsonlWithErrors    = "print-output-jsonl-with-errors"
 	FlagDev                           = "dev"
 	FlagExclude                       = "exclude"
+	FlagExcludePaths                  = "exclude-paths"
 	FlagFile                          = "file"
 	FlagDetectionDepth                = "detection-depth"
 	FlagPruneRepeatedSubdependencies  = "prune-repeated-subdependencies"
--- a/pkg/depgraph/flags.go
+++ b/pkg/depgraph/flags.go
@@ -41,6 +41,7 @@ func getFlagSet() *pflag.FlagSet {
 	flagSet.String(workflow.FlagNugetPkgsFolder, "", "Specify a custom path to the packages folder when using NuGet.")
 	flagSet.Int(workflow.FlagUnmanagedMaxDepth, 0, "Specify the maximum level of archive extraction for unmanaged scanning.")
 	flagSet.Bool(workflow.FlagIncludeProvenance, false, "Include checksums in purl to support package provenance.")
+	flagSet.String(workflow.FlagExcludePaths, "", "Comma-separated paths to exclude from scanning.")
 	flagSet.Bool(workflow.FlagUseSBOMResolution, false, "Use SBOM resolution instead of legacy CLI.")
 	flagSet.Bool(workflow.FlagPrintEffectiveGraph, false, "Return the pruned dependency graph.")
 	flagSet.Bool(workflow.FlagPrintEffectiveGraphWithErrors, false, "Return errors in the pruned dependency graph output.")
--- a/pkg/depgraph/sbom_resolution.go
+++ b/pkg/depgraph/sbom_resolution.go
@@ -86,6 +86,7 @@ func handleSBOMResolutionDI(
 	}
 	allowOutOfSync := !strictOutOfSync
 	exclude := parseExcludeFlag(config.GetString(workflow.FlagExclude))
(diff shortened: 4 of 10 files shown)


pkg/ecosystems/orchestrator/orchestrator.go
```diff
@@ -99,15 +99,26 @@ func LegacyFallback(
 	engine := ictx.GetEngine()
 
 	// Add --print-effective-graph-with-errors for JSONL output with error handling
-	cmdArgs := append([]string(nil), options.Global.RawFlags...)
+	// Extract any user-provided --exclude-relative from RawFlags and merge with processedFiles
+	var existingExcludeRelative string
+	var filteredFlags []string
```
james-snyk:
```suggestion
	filteredFlags := make([]string, 0, len(options.Global.RawFlags))
```
If we can set the size of the array its often a good practice.

pkg/ecosystems/orchestrator/orchestrator.go
```diff
@@ -99,15 +99,26 @@ func LegacyFallback(
...
+	for _, flag := range options.Global.RawFlags {
+		if strings.HasPrefix(flag, "--exclude-relative=") {
+			existingExcludeRelative = strings.TrimPrefix(flag, "--exclude-relative=")
+		} else {
+			filteredFlags = append(filteredFlags, flag)
+		}
+	}
+	cmdArgs := append([]string(nil), filteredFlags...)
 	cmdArgs = append(cmdArgs, "--print-effective-graph-with-errors")
```
james-snyk:
```suggestion
	cmdArgs := append(filteredFlags, "--print-effective-graph-with-errors")
```

pkg/ecosystems/orchestrator/orchestrator.go
```diff
@@ -99,15 +99,26 @@ func LegacyFallback(
...
-	}
-	if len(processedFiles) > 0 {
-		cmdArgs = append(cmdArgs, "--exclude="+strings.Join(processedFiles, ","))
+	if len(processedFiles) > 0 || existingExcludeRelative != "" {
+		parts := make([]string, 0, len(processedFiles)+1)
+		if existingExcludeRelative != "" {
+			parts = append(parts, existingExcludeRelative)
+		}
+		parts = append(parts, processedFiles...)
+		cmdArgs = append(cmdArgs, "--exclude-relative="+strings.Join(parts, ","))
```
james-snyk:
```suggestion
        var parts []string
		if existingExcludeRelative != "" {
			parts = append([]string{existingExcludeRelative}, processedFiles...)
		} else {
			parts = processedFiles
		}
		cmdArgs = append(cmdArgs, "--exclude-relative="+strings.Join(parts, ","))
```
 I think this makes it s bit clear that when we have an existing exclude we just prepend to it, else we just use the processedFiles.

james-snyk (approved):

thomasschafer (approved):

PR #158, opened 2026-04-29
feat: Introduce PluginRegistry with dependency-aware plugin orchestration

james-snyk:

### What this does

Add a new PluginRegistry to manage SCA plugin registration, dependency
resolution, and execution order. This replaces hardcoded plugin invocation
with a flexible, dependency-aware orchestration system.

Key features:
* Explicit dependency declaration at plugin registration time
* Topological sorting using Kahn's algorithm for correct execution order
* Circular dependency detection and validation
* Support for late-added dependencies
* Automatic processed file tracking and exclusion
* Configurable execution modes (single project vs all-projects)

### DIFF
--- a/pkg/ecosystems/orchestrator/registry.go
+++ b/pkg/ecosystems/orchestrator/registry.go
@@ -1,0 +1,172 @@
+package orchestrator
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/rs/zerolog"
+	"github.com/snyk/go-application-framework/pkg/workflow"
+
+	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems"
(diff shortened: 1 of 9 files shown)


pkg/ecosystems/orchestrator/registry.go
```diff
@@ -0,0 +1,172 @@
...
+						queue = append(queue, i)
+					}
+				}
+			}
+		}
+	}
+
+	if len(result) != len(r.entries) {
+		return nil
+	}
```
JamesPatrickGill, line 152: WE should probs error here right? This means there's a cycle?

james-snyk: I made it ignore where cycles happen (they wont get run)

JamesPatrickGill (approved):

PR #160, opened 2026-04-29
feat: bazel jvm & go dep-graph resolvers

gitphill:

# Bazel Dependency Graph Resolver

## Overview

This package builds Snyk dependency graphs from Bazel workspaces. Bazel is a polyglot build system whose dependency model is expressed as a graph of *targets* rather than language-native package coordinates, so vulnerability scanning requires a two-step translation:

1. Ask Bazel itself which targets are reachable from a chosen entry point.
2. Map those Bazel target labels back to the ecosystem coordinates Snyk recognises (Maven `group:artifact:version`, Go `module/path` + version, etc.).

The package is structured around a small `bazelDependencyResolver` interface (`resolver.go`) with one implementation per supported ruleset:

| Language | Ruleset | Lockfile / source-of-truth | Implementation |
| --- | --- | --- | --- |
| JVM (Java, Kotlin, Scala, Android) | [`rules_jvm_external`](https://github.com/bazel-contrib/rules_jvm_external) | `maven_install.json` | `jvm.go` |
| Go | [`rules_go`](https://github.com/bazel-contrib/rules_go) + [`gazelle`](https://github.com/bazel-contrib/bazel-gazelle) | `go.mod` | `go.go` |

The plugin entry point (`plugin.go`) selects a resolver based on CLI flags, walks the discovered targets, and emits one `SCAResult` per target.

## Flags

| Flag | Effect |
| --- | --- |
| `--bazel-jvm` | Enable the `rules_jvm_external` resolver. |
| `--bazel-go` | Enable the `rules_go` resolver. Mutually exclusive with `--bazel-jvm`. |
| `--bazel-target-query` | Override the default Bazel target-discovery query (see below). |
| `--bazel-max-targets` | Maximum number of targets the resolver will process per invocation. Defaults to `1000`; set to `0` to disable the ceiling. |

Without one of `--bazel-jvm` or `--bazel-go`, the plugin no-ops — Bazel projects look like ordinary directories on disk, so there is no reliable heuristic for auto-detection and we keep the trigger explicit.

### Target ceiling

The default queries (`kind('java_binary', //...)` / `kind('go_binary', //...)`) are pre-filtered to deployable entry points, which keeps the result set bounded on most projects. A loose `--bazel-target-query` (e.g. `//...`) can enumerate orders of magnitude more targets and lead to runaway scans. To guard against accidental target explosion, the plugin caps the discovered target count at `1000` and returns an error if exceeded. Raise the ceiling with `--bazel-max-targets=N`, or disable it entirely with `--bazel-max-targets=0` when you genuinely want every target evaluated.

## How resolution works

Both resolvers share the same overall shape:

1. **Build the lookup table.** When the resolver is constructed, it reads the ecosystem-native source-of-truth for versions (`maven_install.json` for JVM, `go.mod` for Go) and indexes it by the *Bazel repository / target name* that the corresponding rules would generate. This means we never have to run `bazel build` to learn versions — they are already pinned in files the user committed.
2. **Find targets.** `findTargets` runs `bazel cquery <query> --output=jsonproto`. The default query is `kind('java_binary', //...)` for JVM and `kind('go_binary', //...)` for Go; both can be overridden via `--bazel-target-query`. Each top-level result becomes a root in its own dep-graph.
3. **Query transitive deps.** For each target, `bazel cquery 'deps(<target>)' --output=jsonproto` returns every rule reachable from that target, along with its label-typed attributes. We extract the language-relevant attributes only (`deps`, `runtime_deps`, `exports` for JVM; `deps`, `embed` for Go) to avoid polluting the graph with toolchain / platform edges that Bazel also reports as "dependencies".
4. **Walk the label graph.** Starting from the root target label, we BFS through the label-to-label edges. Each label is converted into a `PkgInfo` via the lookup table built in step 1; labels that don't match any external repo (i.e. first-party Bazel targets, generated rules, toolchains) are kept verbatim in the graph as intermediate nodes — they have no version, but they preserve the *path* a vulnerable dependency was pulled in through, which is useful when triaging.

All `bazel` subprocesses are dispatched through `query.go`, which uses `--output=jsonproto` and decodes a minimal subset of the [Bazel Build Event Protocol's](https://bazel.build/remote/bep) target message. Only `rule.name` and `rule.attribute` are read; everything else is ignored.

### Lockfile-driven version lookup

`rules_jvm_external` writes a `maven_install.json` lockfile next to the `MODULE.bazel` / `WORKSPACE` file. We read it once at resolver construction time and build a `label → PkgInfo` map.

The trick is that the *keys* in `maven_install.json` are Maven coordinates (`com.google.guava:guava`), while the *labels* Bazel emits in `cquery` output are mangled identifiers (`@maven//:com_google_guava_guava`). `parseArtifactName` re-derives the mangled label from the Maven coordinate by applying `rules_jvm_external`'s own normalisation rules:

- `groupId:artifactId` → `groupId_artifactId`
- `groupId:artifactId:packaging` → `groupId_artifactId` (packaging is dropped)
- `groupId:artifactId:packaging:classifier` → `groupId_artifactId_classifier` (classifier preserved, packaging still dropped)
- Then `.`, `-`, `$` are replaced with `_` to match Bazel's identifier rules.

This mirrors the logic in [`rules_jvm_external/private/rules/artifact.bzl`](https://github.com/bazel-contrib/rules_jvm_external/blob/master/private/rules/artifact.bzl) — if upstream changes its mangling, this function needs to track it.

### Why `maven` as the package manager name

`packageManagerName()` returns `"maven"` rather than something Bazel-specific. Snyk's vulnerability database is keyed by ecosystem, and JVM artifacts coming out of `rules_jvm_external` are ordinary Maven coordinates regardless of which build system pulled them in. Tagging the graph as `maven` lets the existing vulnerability matching pipeline work unchanged.

### Edge attributes

Only `deps`, `runtime_deps`, and `exports` are followed. Bazel rules expose many other label-typed attributes (`plugins`, `data`, `tags`, `srcs`, …) that are not JVM library dependencies in the classpath sense; pulling those into the graph would surface false-positive vulnerable paths.

### `go.mod`-driven version lookup

Go versions live in `go.mod`. We use `golang.org/x/mod/modfile` to parse it and `github.com/bazelbuild/bazel-gazelle/label.ImportPathToBazelRepoName` to compute the same Bazel repository name that `gazelle` / the `bzlmod` `go_deps` extension would generate for each `require`. The result is a `repo-name → PkgInfo` map (e.g. `com_github_spf13_cobra` → `github.com/spf13/cobra @ 1.8.0`).

`replace` directives are honoured when they point at another versioned module — the lookup key stays as the *original* require path (because that is still what gazelle uses to name the Bazel repo), but the value points at the replacement's module path and version. Path-only replaces (e.g. `replace foo => ../local/foo`) have no version and can't be turned into a scannable coordinate, so we fall through to the original require entry.

### Pseudo-version normalisation

Snyk's `gomodules` ecosystem doesn't store Go's semver-prefixed (`v1.2.3`) or pseudo-version (`v0.0.0-20230101000000-abc123def456`) strings verbatim. `normalizeVersion` strips the leading `v` from tagged versions and reduces pseudo-versions to `#<12-char-revision>`, matching the convention used elsewhere in the dep-graph pipeline.

### Sub-package resolution

Go's vulnerability data is keyed by *import path*, not module path — a CVE may apply to `github.com/foo/bar/internal/unsafe` but not to other packages in the same module. Bazel labels carry this information: an external Go target looks like `@com_github_foo_bar//internal/unsafe:unsafe`, where `internal/unsafe` is the in-module package path. `labelToPkgInfo` parses the label, looks up the module by repo name, and concatenates `module + "/" + pkg-path` to produce the final import path. Labels at the module root (`@com_github_foo_bar//:bar`) keep just the module path.

### bzlmod canonical labels

Under bzlmod, `cquery` reports canonical labels like `@@rules_go~~go_deps~com_github_spf13_cobra//cobra:cobra`. The leading `@@…~` / `+` segments encode module-extension provenance; only the trailing apparent-repo segment is what we need for the lookup. The resolver strips everything before the last `~` or `+` in the repo portion before consulting the map.

### Edge attributes

`rules_go` propagates dependencies via `deps` (regular library edges) and `embed` (same-package compilation units stitched together at build time). Both are followed; everything else is ignored.

## First-party and unknown targets

Both resolvers preserve unresolved labels as graph nodes with no version, rather than dropping them. This is deliberate:

- First-party Bazel targets (`//path/to:lib`) are legitimate intermediate nodes in the dep-graph.
- An unrecognised external repo could indicate a missing lockfile entry, a typo in `parseArtifactName`'s mangling rules, or a brand-new ruleset we haven't taught the resolver about. Keeping the node visible makes those cases triagable instead of silently lossy.

The downside is that "phantom" nodes without versions cannot be matched against the vuln database — they show up in the graph but contribute nothing to a scan. The expectation is that real ecosystem coordinates eventually appear *below* them as transitive children.

## Why `cquery` and not `query`

`bazel query` works on the unconfigured target graph and doesn't see select-resolved or platform-conditional deps. `bazel cquery` works on the *configured* target graph, which is what actually gets built. For accurate vulnerability reporting we need the latter — e.g. an Android `cc_library` that selects between platform-specific deps via `select({})` should report whichever branch is active in the user's build configuration.

The trade-off: `cquery` requires the workspace to be loadable (toolchains resolved, repository rules fetchable). Workspaces that can't load will fail target discovery; a `query`-based fallback for partial graph extraction is something we may revisit if it becomes a recurring blocker.

## Integration tests

Integration tests live alongside the unit tests in this package and are gated behind environment variables so they don't run in the default `go test ./...` pass — they need a working `bazel` on `PATH` and (for the Android JVM fixture) `ANDROID_HOME` pointing at an SDK.

```sh
make test-bazel-jvm-integration   # BAZEL_JVM_INTEGRATION_TESTS=1
make test-bazel-go-integration    # BAZEL_GO_INTEGRATION_TESTS=1
```

Fixtures live under `pkg/ecosystems/testdata/fixtures/bazel/` they are working examples of Bazel projects.


pkg/ecosystems/orchestrator/registry.go
```diff
@@ -32,6 +33,10 @@ func NewDefaultPluginRegistry() (*PluginRegistry, error) {
 		plugins: make([]ecosystems.SCAPlugin, 0),
 	}
 
+	// bazel
+	if err := r.register(bazel.Plugin{}); err != nil {
+		return nil, fmt.Errorf("failed to register bazel plugin: %w", err)
+	}
```
JamesPatrickGill: @james-snyk can help here, but I think we need to make sure **every** other plugin depends on this so it runs first. 

Can do this by adding it as a dep to the others - probably most likely to stay consistent that way

gitphill: I think it's only a problem when there are shared manifests; which is not the case in JVM rules, for example maven_install.json != pom.xml; but it is the case for go, there's a go.mod. So bazel is a dependency of any go scanner we add, but not of maven or gradle or sbt.

pkg/ecosystems/bazel/plugin.go
```diff
@@ -0,0 +1,107 @@
...
+
+const (
+	pluginName = "bazel"
+	// defaultMaxTargets caps target enumeration to protect Snyk from runaway
+	// scans on monorepos that match an overly broad --bazel-target-query. The
+	// ceiling can be raised via --bazel-max-targets, or set to 0 to disable.
+	defaultMaxTargets = 1000
+)
+
+type Plugin struct{}
```
mcombuechen, line 21: prolly still needs a 
```go
var _ ecosystems.SCAPlugin = new(Plugin)
```

.circleci/continue-config.yml
```diff
@@ -6,7 +6,10 @@ orbs:
 parameters:
   go_version:
     type: string
-    default: "1.26.3"
+    default: "1.25.7"
```
calhar-snyk: Given `go 1.26.3` in the `go.mod`, we probably shouldn't be downgrading here.

gitphill: ah yep, i wrote this pre upgrade it seems; aligning with main

pkg/ecosystems/bazel/plugin.go
gitphill: 👍 is that how go knows the plugin signs up to the SCAPlugin interface?

mcombuechen: yes, it's some compile time check to make sure it properly implements the interface. looks a bit funky, but kinda neat as a sanity check.

pkg/ecosystems/bazel/plugin.go
```diff
@@ -0,0 +1,107 @@
...
+	dir string,
+	options *ecosystems.SCAPluginOptions,
+) (*ecosystems.PluginResult, error) {
+	if log == nil {
+		log = logger.Nop()
+	}
+
+	resolver, err := newResolverFromOptions(dir, options)
+	if err != nil {
+		if errors.Is(err, errNoBazelOptionFound) {
```
calhar-snyk, line 42: Nitpick: I don't necessarily love this as a flow control mechanism. It feels to me like the no option case should return either a NoopResolver, or `(nil, nil)`

pkg/ecosystems/bazel/go.go
```diff
@@ -0,0 +1,271 @@
...
+func (r *goResolver) packageManagerName() string {
+	return "gomodules"
+}
+
+// processedFiles reports go.mod and go.sum as consumed so that the legacy CLI does not
+// re-scan it after the Bazel resolver has already produced a dep-graph from it.
+func (r *goResolver) processedFiles() []string {
+	return []string{goModFilename, goSumFilename}
+}
+
```
cjheppell, line 130: Is this correct? 

We're running within the context of a directory - shouldn't the processed files be returning a filepath for the go.mod and go.sum relative to that directory? Not just the filenames?

gitphill: oh not sure @JamesPatrickGill @james-snyk what's the convention here? paths or names

pkg/ecosystems/bazel/plugin.go
gitphill: ok yea, i'll use the (nil, nil) pattern

pkg/ecosystems/testdata/fixtures/bazel/rules-go-0.52.0/basic-gazelle/go_integration_test.snap
```diff
@@ -0,0 +1,244 @@
...
+      "id": "k8s.io/klog/v2@2.80.1",
+      "info": {
+       "name": "k8s.io/klog/v2",
+       "version": "2.80.1"
+      }
+     },
+     {
+      "id": "//pkg/roll:roll@",
+      "info": {
+       "name": "//pkg/roll:roll"
```
calhar-snyk, line 165: Things like this seem like they should either be getting normalized to a more useful name or omitted as dependencies.

pkg/ecosystems/testdata/fixtures/bazel/rules-jvm-external-6.10/spring_boot/plugin_integration_test.snap
calhar-snyk, line 1: Why both `jvm_integration_test.snap` and `plugin_integration_test.snap`? There seems to be very little difference between the two, while the go fixtures only have `go_integration_test.snap`

pkg/ecosystems/bazel/go.go
```diff
@@ -0,0 +1,271 @@
...
+func (r *goResolver) packageManagerName() string {
+	return "gomodules"
+}
+
+// processedFiles reports go.mod and go.sum as consumed so that the legacy CLI does not
+// re-scan it after the Bazel resolver has already produced a dep-graph from it.
+func (r *goResolver) processedFiles() []string {
+	return []string{goModFilename, goSumFilename}
+}
```
JamesPatrickGill, line 129: I know this might not be ready right now, but for --all-projects this needs to be the exact files used to produce the depgraph in this plugin, not just names.

This would just always ignore root go projects even if you aren't using bazel

pkg/ecosystems/bazel/plugin.go
```diff
@@ -0,0 +1,109 @@
...
+	if err := checkTargetLimit(len(targets), options); err != nil {
+		return nil, err
+	}
+
+	var results []ecosystems.SCAResult
+	for _, target := range targets {
+		graph, err := resolver.buildDepGraph(ctx, target)
+		if err != nil {
+			log.Error(ctx, "failed to build graph for bazel target", logger.Attr("target", target), logger.Err(err))
+		}
```
JamesPatrickGill, line 66: This error block doesn't continue or anything

pkg/ecosystems/bazel/plugin.go
calhar-snyk: Probably worth having similar statements for the go and jvm resolvers as well to ensure they're implementing the interface

pkg/ecosystems/bazel/go.go
```diff
@@ -0,0 +1,271 @@
...
+			continue
+		}
+		lookup[repoName] = depgraph.PkgInfo{
+			Name:    modPath,
+			Version: normalizeVersion(modVersion),
+		}
+	}
+
+	return lookup, nil
+}
```
JamesPatrickGill, line 101: For this gazelle stuff I'm not seeing a fixture that exercises a versioned replace? The unit tests own both sides of the logic

pkg/ecosystems/bazel/go.go
```diff
@@ -0,0 +1,271 @@
...
+		if result.Target == nil || result.Target.Rule == nil {
+			continue
+		}
+		if n := result.Target.Rule.Name; n != "" {
+			targets = append(targets, n)
+		}
+	}
+
+	return targets, nil
+}
```
JamesPatrickGill, line 153: nit: You could combine with JVM

pkg/ecosystems/bazel/jvm.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	parts := strings.Split(name, ":")
+	var l string
+	var n string
+
+	switch len(parts) {
+	case 1:
+		l = parts[0]
+		n = parts[0]
```
JamesPatrickGill, line 97: question: Can this ever be just 1?

pkg/ecosystems/bazel/query.go
```diff
@@ -0,0 +1,46 @@
...
+					Name            string   `json:"name"`
+					StringListValue []string `json:"stringListValue"`
+				} `json:"attribute"`
+			} `json:"rule"`
+		} `json:"target"`
+	} `json:"results"`
+}
+
+func bazelQuery(ctx context.Context, dir, query string) (*queryResults, error) {
+	cmd := exec.CommandContext(ctx, bazelCommand, "cquery", query, "--output=jsonproto")
```
JamesPatrickGill, line 30: question: you mentioned queries can go exponential before, do we need to add a timeout inside this function that isn't caller controlled?

pkg/ecosystems/bazel/go.go
gitphill: ah yea i reckon it's path; that's what CLI in Typescript land wants to know about

pkg/ecosystems/testdata/fixtures/bazel/rules-go-0.52.0/basic-gazelle/go_integration_test.snap
gitphill: i wanted to use the Bazel label name as used; i think that's what's most useful to a developer to reason over the vulnerable path

pkg/ecosystems/testdata/fixtures/bazel/rules-jvm-external-6.10/spring_boot/plugin_integration_test.snap
gitphill: oh that's a previous run i think, when the test was named differently; i'll remove

pkg/ecosystems/bazel/go.go
gitphill: 👍 yep i'll use absolute paths

pkg/ecosystems/bazel/go.go
gitphill: Yea good point; I'll see if I can find one. There's are only a one example in the rules go repo, but I can make one up

gitphill: added `gazelle-version-replace`

pkg/ecosystems/bazel/go.go
gitphill: ok they're paths now

pkg/ecosystems/bazel/plugin.go
gitphill: oh apparently nil, nil is not the way https://github.com/Antonboom/nilnil

pkg/ecosystems/bazel/plugin.go
gitphill: 👍 adding a continue

pkg/ecosystems/bazel/go.go
gitphill: meh, little bit of duplication is alright; perhaps if we see this again

pkg/ecosystems/bazel/jvm.go
gitphill: not according to the rule code https://github.com/bazel-contrib/rules_jvm_external/blob/master/private/coursier_utilities.bzl#L45 but we can be defensive?

pkg/ecosystems/bazel/query.go
gitphill: timeout maybe more tricky to limit; i'm not sure what normal looks like on large enterprise mono repos

pkg/ecosystems/bazel/jvm.go
JamesPatrickGill: dead branch then? Maybe we clean up in a follow

pkg/ecosystems/bazel/jvm.go
```diff
@@ -0,0 +1,258 @@
...
+	for k, v := range lockfile.Artifacts {
+		a := parseArtifactName(k)
+		if a.label == "" {
+			continue
+		}
+		lookup[a.label] = depgraph.PkgInfo{
+			Name:    a.name,
+			Version: v.Version,
+		}
+	}
```
JamesPatrickGill, line 70: I think we can hit collisions here leading to non deterministic behaviour as the range over a map is non deterministic. 

Path would be we get two artifacts that are the same as we are dropping the packaging part of the label. These duplicates may then be processed in different orders on different tests

pkg/ecosystems/bazel/go.go
```diff
@@ -0,0 +1,274 @@
...
+		modulePath string
+		version    string
+	}
+	replaces := make(map[string]replacement, len(parsed.Replace))
+	for _, r := range parsed.Replace {
+		if r.New.Version == "" {
+			continue
+		}
+		replaces[r.Old.Path] = replacement{modulePath: r.New.Path, version: r.New.Version}
+	}
```
JamesPatrickGill, line 79: This globally replaces but just on name. In go my understanding is that replace also works on version so we can't just replace universally, i.e:

```go
replace foo v1.0.0 => bar v1.0.1
```

just applied to v1.0.0 but your code looks like it steamrolls every version

pkg/ecosystems/bazel/jvm.go
```diff
@@ -0,0 +1,258 @@
...
+	for _, result := range output.Results {
+		if result.Target == nil || result.Target.Rule == nil {
+			continue
+		}
+		if n := result.Target.Rule.Name; n != "" {
+			targets = append(targets, n)
+		}
+	}
+
+	return targets, nil
```
JamesPatrickGill, line 156: This, with a similar case in the go version of this loop, might cause duplicates as the bazel query returns targets and i think the same rule can appear under multiple configurations. So we will have two exactly the same binaries and results being produced.

Probs an edge case but worth shouting out

JamesPatrickGill (approved):
Added three comments to address as @gitphill iterates. 

Merging now so we can continue improving outside of this large PR.

JamesPatrickGill:
> Added three comments to address as @gitphill iterates.
> 
> Merging now so we can continue improving outside of this large PR.

After merge we also need to address the go mod processed files, they aren't actually mapping to files used to create them, so duplicate finds is a possibility.

Okay for now as we are just in test but we should iterate on in small PRs

PR #159, opened 2026-04-29
feat: [CMPA-473] Add integration testing for gradle

calhar-snyk:
- [ ] Tests written and linted [ℹ︎](https://github.com/snyk/general/wiki/Tests)
- [ ] Documentation written [ℹ︎](https://github.com/snyk/general/wiki/Documentation)
- [ ] Commit history is tidy [ℹ︎](https://github.com/snyk/general/wiki/Git)

### Summary

Add integration testing infrastructure for Gradle ecosystem support, enabling real Gradle invocations to validate dependency graph generation across different project scenarios.

### Key Changes

**Integration Test Framework**
- Introduced `plugin_integration_test.go` with test harness using `integration,gradle` build tags
- Added fixture management system with snapshot generation and updating via `UPDATE_FIXTURES=1`
- Implemented metadata extraction utilities for test fixtures and Gradle runtime version detection

**Test Coverage & Scenarios**  
- Created test fixtures covering key Gradle use cases:
  - **Classifiers**: JAR classifiers and dependency variants
  - **Dynamic versions**: Version range resolution and constraints
  - **Kotlin DSL**: `.gradle.kts` build script support  
  - **Multi-module projects**: Complex project structures with target file scanning
- Each fixture includes Gradle project structure with expected output snapshots

**CI/CD Integration**
- Updated CircleCI configuration to run integration tests across matrix of `(gradle, jdk)` combinations: `[(6,11), (7,17), (8,21), (9,25)]`
- Added path-based test triggering when Gradle-related files are modified
- Configured official `gradle:<X>-jdk<Y>` Docker images for test environments

PR #164, opened 2026-05-05
feat: Adds `ResolverMetadata` to SCAResult containing optional details about resolver execution

calhar-snyk:

### What this does

Adds a `ResolverMetadata` struct to the `SCAResult`. This maps to the legacy plugin `meta` field and contains details about the resolver that may be useful for debugging/tracking/metrics:
- Name
- Underlying versions
- etc.

### DIFF
--- a/pkg/ecosystems/README.md
+++ b/pkg/ecosystems/README.md
@@ -34,9 +34,10 @@ The `SCAResult.DepGraph` field uses the standard Snyk dependency graph format fr
 import "github.com/snyk/dep-graph/go/pkg/depgraph"
 
 type SCAResult struct {
-    DepGraph *depgraph.DepGraph `json:"depGraph,omitempty"`
-    Metadata Metadata           `json:"metadata"`
-    Error    error              `json:"error,omitempty"`
+    DepGraph          *depgraph.DepGraph         `json:"depGraph,omitempty"`
+    ProjectDescriptor identity.ProjectDescriptor `json:"projectDescriptor"`
+    ResolverMetadata  *ResolverMetadata          `json:"meta,omitempty"`
+    Error             error                      `json:"error,omitempty"`
 }
 ```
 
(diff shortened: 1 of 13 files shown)


pkg/ecosystems/plugin_interface.go
```diff
@@ -9,11 +9,17 @@ import (
 	"github.com/snyk/cli-extension-dep-graph/pkg/identity"
 )
 
+type ResolverMetadata struct {
+	PluginName       string            `json:"pluginName"`
+	VersionBuildInfo map[string]string `json:"versionBuildInfo"`
```
james-snyk:
```suggestion
	PluginName       string            `json:"pluginName,omitempty"`
	VersionBuildInfo map[string]string `json:"versionBuildInfo,omitempty"`
```

pkg/ecosystems/python/pipenv/plugin.go
```diff
@@ -226,6 +226,12 @@ func (p Plugin) buildDepGraphFromPipfile(
 				RootComponentName: rootName,
 			},
 		},
+		ResolverMetadata: &ecosystems.ResolverMetadata{
+			PluginName: PluginName,
+			VersionBuildInfo: map[string]string{
+				"pythonVersion": pythonVersion,
```
james-snyk: Should we create a enums for all the possible key values (pythonVersion, gradleVersion, javaVersion, generatedAt)?

Dont feel strongly about this but thought i would raise the idea.

calhar-snyk: I don't see any harm in having these codified. It's not like more keys can't be trivially added.

tonidevine1:
Maybe we should add some tests to confirm that the meta field is populated correctly in each plugin?

tonidevine1:
> > Maybe we should add some tests to confirm that the meta field is populated correctly in each plugin?
> 
> I'll be adding some integration tests in #159 for gradle integration tests. I'm not necessarily convinced about the value other ones yet -- seems a bit asserting trivial setting/getters to me. But I'm not strongly against adding some

Yeah that makes sense. I do agree about not adding pointless tests, I was more just thinking to protect this if there are any changes made in the future to refactor the plugin's result.

james-snyk (approved):

pkg/ecosystems/README.md
```diff
@@ -133,23 +134,76 @@ depGraph := builder.Build()
 
 ```go
 type SCAResult struct {
```
calhar-snyk, line 136: The duplicate reporting of the `SCAResult` here and on `L36` are pre-existing. I'm not convinced about the need, but figured I'd update both.

PR #165, opened 2026-05-06
refactor: migrate logic from os-flows to dep-graph

mcombuechen:
### What this does

This is Step 1 of a larger refactor that aims to move ecosystem-specific logic from cli-extension-os-flows to cli-extension-dep-graph, specifically this: https://github.com/snyk/cli-extension-os-flows/blob/d362d2b49ac88c21a69df7efcb849eab656fe5bb/internal/common/depgraph.go#L49-L76

It moves two logical branches over:
* uv resolution when a uv.lock file is present and the FF is on (aka. "use-sbom-resolution")
* resolving dep-graphs via the orchestrator if new FF `unified-test-api-os-cli` is turned on

The intention is to remove logic from the os-flows domain that actually falls into the ecosystems domain.

Step 2 will be a PR in cli-extension-os-flows to remove these logical branches and to switch to `engine.InvokeWithConfig()` as the single entrypoint to the dep-graph extension.

Step 3 will be an extension upgrade in the CLI with new acceptance tests.

### DIFF
--- a/pkg/depgraph/callback.go
+++ b/pkg/depgraph/callback.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/snyk/cli-extension-dep-graph/internal/legacycli"
 	"github.com/snyk/cli-extension-dep-graph/internal/workflow"
+	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems/python/uv"
 )
 
 type ResolutionHandlerFunc func(ctx gafworkflow.InvocationContext, config configuration.Configuration, logger *zerolog.Logger) ([]gafworkflow.Data, error)
@@ -17,9 +18,36 @@ func callback(ctx gafworkflow.InvocationContext, _ []gafworkflow.Data) ([]gafwor
 
 	logger.Print("DepGraph workflow start")
 
-	if config.GetBool(workflow.FlagUseSBOMResolution) {
+	if shouldUseOrchestratorResolution(config) {
+		return handleOrchestratorResolution(ctx, config, logger)
+	}
+
+	if shouldUseSBOMResolution(config, logger) {
 		return handleSBOMResolution(ctx, config, logger)
 	}
 
 	return legacycli.HandleLegacyResolution(ctx, config, logger) //nolint:wrapcheck // must return unwrapped so os-flows can detect and render ErrorCatalog.
+}
+
+func shouldUseSBOMResolution(config configuration.Configuration, logger *zerolog.Logger) bool {
+	if !uv.HasLockFile(
+		config.GetString(configuration.INPUT_DIRECTORY),
+		config.GetString(workflow.FlagFile),
+		config.GetBool(workflow.FlagAllProjects),
+		logger,
+	) {
+		return false
+	}
+
+	if !config.GetBool(workflow.FeatureFlagUvCLI) {
+		return false
+	}
+
+	logger.Info().Msg("uv.lock found and uv feature flag enabled, using SBOM resolution")
+
+	return true
+}
+
+func shouldUseOrchestratorResolution(config configuration.Configuration) bool {
+	return config.GetBool(workflow.FeatureFlagUseUnifiedTestAPIForOSCliTest)
 }
--- a/internal/workflow/constants.go
+++ b/internal/workflow/constants.go
@@ -5,6 +5,9 @@ const (
 	ContentTypeJSONL       = "application/jsonl"
 	LegacyCLIWorkflowIDStr = "legacycli"
 	ContentLocationKey     = "Content-Location"
+
+	FeatureFlagUvCLI                         = "internal_snyk_cli_uv_enabled"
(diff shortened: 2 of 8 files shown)


pkg/depgraph/callback.go
```diff
@@ -17,9 +18,33 @@ func callback(ctx gafworkflow.InvocationContext, _ []gafworkflow.Data) ([]gafwor
...
+
+	if shouldUseSBOMResolution(config, logger) {
 		return handleSBOMResolution(ctx, config, logger)
 	}
 
 	return legacycli.HandleLegacyResolution(ctx, config, logger) //nolint:wrapcheck // must return unwrapped so os-flows can detect and render ErrorCatalog.
 }
+
+func shouldUseSBOMResolution(config configuration.Configuration, logger *zerolog.Logger) bool {
+	if !config.GetBool(workflow.FeatureFlagUvCLI) {
```
thomasschafer: Should we reorder this and the lockfile check? Jedi wanted us to minimise FF calls, so in the os-flows extension we check for the lockfile before checking the flag

pkg/depgraph/callback.go
```diff
@@ -17,9 +18,33 @@ func callback(ctx gafworkflow.InvocationContext, _ []gafworkflow.Data) ([]gafwor
 
 	logger.Print("DepGraph workflow start")
 
-	if config.GetBool(workflow.FlagUseSBOMResolution) {
+	if shouldUseOrchestratorResolution(config) {
+		return handleOrchestratorResolution(ctx, config, logger)
```
thomasschafer, line 22: Why do we need this separate branch? We also seem to be calling the same function directly in the os-flows extension, so we seem to have (at least) 3 ways to resolve dep-graphs in the CLI now based on which flags are enabled. Why do we need these 3 variants? Why not have everything go through `handleSBOMResolution`, and just enable or disable plugins based on flags?

pkg/depgraph/callback.go
mcombuechen: ah, i didn't have this context. yeah good call, i will switch it around!

pkg/depgraph/callback.go
mcombuechen: all the branches in the os-flows extension (including the direct call to orchestrator) will go away. I opened a draft PR here which illustrates this: https://github.com/snyk/cli-extension-os-flows/pull/234

i still wanted to retain the 3 branches **1st uv** (if applicable), **2nd orchestrator** (if applicable), **3rd legacy fallback**. It's not more branches than before, just that they have moved from os-flows to this extension. again, the draft PR should illustrate this.

thomasschafer: That makes sense, really my question was about why the os-flows extension had these additional branches e.g. the direct call to the orchestrator in the first place, rather than going via the dep-graph extension?

pkg/depgraph/callback.go
mcombuechen: done

pkg/depgraph/callback.go
thomasschafer: @mcombuechen let me know if you want to discuss this - I'm just concerned that `pkg/depgraph/orchestrator_resolution.go` will become the *fourth* orchestrator we have (`sbom_resolution.go`, `orchestrator.go`, `registry.go` and now `orchestrator_resolution.go`) - from my understanding `sbom_resolution.go` is the only one actually used by customers so would be great to trim things down

mcombuechen: os-flows had the branching logic because (a) is already evaluated a new FF that would route to the orchestrator, which was lazy implementation, so it was easy to read the FF and invoke the orchestrator, but more importantly it was not possible to call the orchestrator from within the `callback()` function (in `pkg/depgraph/workflow.go`) because it would've introduced an import cycle. This has been fixed in a previous PR, and we can now safely import and invoke the orchestrator in `callback()`.

So simplistically, it's like this: 

**BEFORE**

```
os-flows:
  -> if FF enableUvCLI, lockfile exists
    -> set config "use-sbom-resolution"
  -> if FF unified-test-api-os-cli
    -> call orchestrator.ResolveDepgraphs(), stop
  finally, invoke depgraph workflow

dep-graph:
  -> if config "use-sbom-resolution"
    -> enter sbom/uv orchetrator, stop
  invoke legacy workflow
```

**AFTER**

```
os-flows:
  invoke depgraph workflow
  
dep-graph:
  -> if FF enableUvCLI, lockfile exists
    -> enter sbom/uv orchetrator, stop
  -> if FF unified-test-api-os-cli
    -> call orchestrator.ResolveDepgraphs(), stop
  finally, invoke legacy workflow
```

This maybe illustrates how we remove many logical branches from os-flows, and push them down into cli-extension-dep-graph. No 4th orchestrator as far as I can tell.

thomasschafer (approved):

james-snyk (approved):

james-snyk (changes requested):
Might need a talk about this again as doesn't fully align with the direction we want to go down.

mcombuechen:
Closing this PR for now, maybe we can pick up the GAF Workflow discussion at a later point.

PR #166, opened 2026-05-07
refactor: wrap legacy CLI execution as an SCAPlugin

thomasschafer:
- [ ] Documentation written [ℹ︎](https://github.com/snyk/general/wiki/Documentation) - n/a

### What this does

Wraps legacy CLI execution as an `SCAPlugin` so the SBOM-resolution path and the plugin-loop orchestrator both invoke it through the same interface as `uv`, `bun` etc.

The dep-graph extension's two entry-points (the orchestrator's `ResolveDepgraphs` loop and `pkg/depgraph/sbom_resolution.go`) previously called legacy through bespoke shims:
- `legacy.NewLegacyResolver(ictx, processedFiles)` (orchestrator path)
- `executeLegacyWorkflow` -> `processLegacyData` (SBOM-resolution path)

Both are replaced by a single `legacy.NewPlugin(ictx)` that satisfies `ecosystems.SCAPlugin`.

`pkg/ecosystems/legacy/identity.go` and its per-ecosystem inference table are removed, as this is just emulating `TargetFileFromPlugin`, so we can use the latter directly instead.


pkg/ecosystems/orchestrator/registry.go
```diff
@@ -58,7 +58,6 @@ func (r *PluginRegistry) ResolveDepgraphs(ictx workflow.InvocationContext, dir s
 
 		ctx := ictx.Context()
 		enhancedLogger := ictx.GetEnhancedLogger()
-		processedFiles := make([]string, 0)
```
thomasschafer, line 56: Wasn't used given `opts.WithExclude(files)` below, which passes exclusions between plugins. The legacy CLI will need a new flag like `--exclude-paths` e.g. see https://github.com/snyk/cli-extension-dep-graph/pull/152

thomasschafer:
TODO: I'm seeing 500 errors on this branch when running with `--all-projects`

pkg/depgraph/sbom_resolution.go
```diff
@@ -358,7 +294,9 @@ func combineWorkspaceResultsAsJSONL(logger *zerolog.Logger, results []ecosystems
 	targetFile := results[0].ProjectDescriptor.GetTargetFile()
 	workflowData.SetMetaData(workflow.ContentLocationKey, targetFile)
 	workflowData.SetMetaData(workflow.MetaKeyNormalisedTargetFile, targetFile)
-	workflowData.SetMetaData(workflow.MetaKeyTargetFileFromPlugin, targetFile)
```
thomasschafer, line 346: Check if this is needed in registry.go

pkg/identity/project.go
```diff
@@ -16,6 +16,24 @@ type ProjectIdentity struct {
...
 	// RootComponentName specifies the component's name that is at the root of the project
 	RootComponentName string `json:"rootComponentName,omitempty"`
+	// Legacy holds fields that exist solely to bridge legacy CLI quirks. New plugin
+	// authors should leave this nil.
+	Legacy *LegacyIdentity `json:"legacy,omitempty"`
+}
+
+// LegacyIdentity carries fields that exist solely to bridge quirks of the legacy Snyk CLI.
+// New plugins do not need to populate this.
+type LegacyIdentity struct {
```
thomasschafer: TODO: confirm where exactly the flow breaks downstream when we don't have this

pkg/depgraph/sbom_resolution.go
thomasschafer: Not needed as `registry.go` doesn't try to map back to workflow data

pkg/ecosystems/orchestrator/registry.go
thomasschafer: Now converted to `WithExcludePaths`

pkg/identity/project.go
thomasschafer: Here are the metadata fields set by the legacy CLI:

| Ecosystem | Case | normalisedTargetFile and Content-Location | targetFileFromPlugin | target [^t] | workspace |
|---|---|---|---|---|---|
| **npm** [^1] | workspace root | `<manifest path>` | _absent_ | `{}` | `{"type":"npm"}` |
| **npm** | workspace child | `<manifest path>` | _absent_ | `{}` | `{"type":"npm"}` |
| **npm** | non-workspace | `<manifest path>` | _absent_ | `{}` | _absent_ |
| **yarn** [^2] | workspace root | `<manifest path>` | _absent_ | `{}` | `{"type":"yarn"}` |
| **yarn** | workspace child | `<manifest path>` | _absent_ | `{}` | `{"type":"yarn"}` |
| **yarn** | non-workspace (`yarn.lock` only) | `<manifest path>` | _absent_ | `{}` | _absent_ |
| **pnpm** [^2] | workspace root | `<manifest path>` | _absent_ | `{}` | `{"type":"pnpm"}` |
| **pnpm** | workspace child | `<manifest path>` | _absent_ | `{}` | `{"type":"pnpm"}` |
| **gomodules** [^3] | `go.mod` | `<manifest path>` | `<manifest path>` | `{}` | _absent_ |
| **gradle** [^4] | `build.gradle.kts` | `<manifest path>` | `<manifest path>` | `{}` | _absent_ |
| **gradle** [^5] | `build.gradle` (Groovy) | `<manifest path>` | _absent_ | `{}` | _absent_ |
| **maven** [^6] | `pom.xml` | `<manifest path>` | _absent_ | `{}` | _absent_ |
| **rubygems** [^7] | `Gemfile.lock` | `<manifest path>` | _absent_ | `{}` | _absent_ |
| **sbt** [^8] | `build.sbt` | `<manifest path>` | _absent_ | `{}` | _absent_ |
| **composer** [^3] | `composer.lock` | `<manifest path>` | `<manifest path>` | `{}` | _absent_ |
| **nuget** [^9] | `project.assets.json` | `<manifest path>` | `<manifest path>` | `{}` | _absent_ |
| **paket** [^3] | `paket.dependencies` | `<manifest path>` | `<manifest path>` | `{}` | _absent_ |
| **pip** [^10] | `Pipfile` | `<manifest path>` | `<manifest path>` | `{}` | _absent_ |
| **pip** [^11] | `requirements.txt` | `<manifest path>` | _absent_ | `{}` | _absent_ |
| **poetry** [^3] | `pyproject.toml` (poetry-managed) | `<manifest path>` | `<manifest path>` | `{}` | _absent_ |
| **cocoapods** [^3] | `Podfile` | `<manifest path>` | `<manifest path>` | `{}` | _absent_ |
| **swift** [^12] | `Package.swift` | `swift-projPackage.swift` ⚠️ | `swift-projPackage.swift` ⚠️ | `{}` | _absent_ |

[^1]: snyk-nodejs-plugin omits `plugin.targetFile` for all npm cases.
[^2]: Same plugin as npm (snyk-nodejs-plugin).
[^3]: Plugin sets `plugin.targetFile`.
[^4]: snyk-gradle-plugin sets it for `.kts` only.
[^5]: snyk-gradle-plugin omits for non-`.kts`.
[^6]: snyk-mvn-plugin omits.
[^7]: snyk-gemfile-plugin omits.
[^8]: snyk-sbt-plugin omits.
[^9]: Plugin sets `plugin.targetFile`; also sets `targetRuntime` (e.g. `"net6.0"`) — the only ecosystem in this sample that does.
[^10]: snyk-python-plugin sets for `Pipfile`.
[^11]: snyk-python-plugin omits for `requirements.txt`.
[^12]: Plugin sets `plugin.targetFile`. **Note**: legacy CLI emits the path without a `/` separator between dir and filename (`swift-projPackage.swift` instead of `swift-proj/Package.swift`) — a legacy-CLI quirk specifically in the swift plugin's path handling.
[^t]: The cells show `{}` because this run was against a directory that wasn't a Git checkout. When the scan runs inside a Git repository with a configured remote, the legacy CLI auto-detects it and `target` becomes `{"remoteUrl":"<url>"}` (e.g. `{"remoteUrl":"http://github.com/example/test.git"}`) for every successful entry. Container scans carry a `ContainerTarget` shape instead (image name, digest, etc.). The `--remote-repo-url` CLI flag does **not** populate this field — it's used downstream for project association in the API request, not in the per-result `target` payload.

Given this, we need to unconditionally set `"Content-Location"` and `"normalisedTargetFile"` as the target file, and `"targetFileFromPlugin"` is set conditionally. Previously `getProjectTargetFileBasedOnType` did this manually, but now we just pass the actual value out of the plugin for this key, along with `target` which was not previously returned.

internal/legacycli/errors.go
```diff
@@ -14,6 +14,21 @@ import (
...
 var ErrNoDepGraphsFound = errors.New("no depgraphs found")
 
+// ExitCoder allows checking for exit codes without depending on a concrete error type.
+type ExitCoder interface {
+	ExitCode() int
+}
+
+// IsExitCode3 reports whether the error chain contains an error with exit code 3.
+// Exit code 3 from the legacy CLI means "no projects found to test".
+func IsExitCode3(err error) bool {
```
mcombuechen: Just a nit, but maybe we can rename this to something more descriptive like `IsNoProjectFoundError`; because I think this would give it more context around the callsite where it is used. Let me know what you think.

pkg/ecosystems/legacy/plugin.go
```diff
@@ -10,245 +10,120 @@ import (
...
-
-	// Invoke legacy CLI workflow
-	legacyData, err := ictx.GetEngine().InvokeWithConfig(legacyCLIWorkflowID, legacyConfig)
+	depGraphs, err := legacycli.InvokeLegacy(l.ictx, legacyConfig, l.ictx.GetEnhancedLogger())
 	if err != nil {
-		if strings.Contains(err.Error(), "No supported files found") && len(ignores) > 0 {
-			return nil, nil
+		if errors.Is(err, legacycli.ErrNoDepGraphsFound) || legacycli.IsExitCode3(err) {
+			log.Debug(ctx, "No projects found in legacy CLI call")
+			return &ecosystems.PluginResult{}, nil
```
mcombuechen: can't we return `nil, nil` here? I find the empty `&ecosystems.PluginResult{}` problematic because it suggests we did find _something_, it's not really accurately describing the "no projects" result.

pkg/ecosystems/orchestrator/orchestrator.go
```diff
@@ -49,8 +49,21 @@ func resolveLegacy(
 	opts *ecosystems.SCAPluginOptions,
 	ignores []string,
 ) (*ecosystems.PluginResult, error) {
+	if len(ignores) > 0 && !opts.Global.AllProjects {
+		return &ecosystems.PluginResult{}, nil
+	}
```
mcombuechen: I'm not sure I understand the logic here. If we have ignores (which can be set by the user if I'm not mistaken) we bail out? That does not seem right.

Also, we probably want to avoid `&ecosystems.PluginResult{}` (empty result) here as well and prefer a `nil`.

pkg/ecosystems/legacy/plugin.go
```diff
@@ -257,26 +132,32 @@ func depGraphOutputToSCAResult(ictx workflow.InvocationContext, output *parsers.
...
+			// upstream error is likely more user-meaningful than our parse failure.
+			if result.Error == nil {
+				result.Error = fmt.Errorf("failed to unmarshal depgraph: %w", err)
+			}
+			return result
 		}
 
 		result.DepGraph = &dg
 		result.ProjectDescriptor.Identity.ProjectType = dg.PkgManager.Name
-		result.ProjectDescriptor.Identity.TargetFile = getProjectTargetFileBasedOnType(dg.PkgManager.Name, output.NormalisedTargetFile, output.Workspace != nil)
```
thomasschafer, line 274: Not needed as we have `TargetFile: output.TargetFileFromPlugin` above, which is exactly what `getProjectTargetFileBasedOnType` was trying to emulate

internal/legacycli/errors.go
thomasschafer: Good point, done!

pkg/ecosystems/orchestrator/orchestrator.go
thomasschafer: On the first point: `ignores` is just `pythonResult.ProcessedFiles` ([here](https://github.com/snyk/cli-extension-dep-graph/blob/8ab6045/pkg/ecosystems/orchestrator/orchestrator.go#L27)), and so if we have ignores then we have found files, so we can stop going, unless we are using `AllProjects`. I know that this whole file is going to be deleted soon so I debated deleting it in this PR, but I thought that that would be overstepping so better to fix it while here than leave it in a broken state

pkg/ecosystems/legacy/plugin.go
thomasschafer: Updated

pkg/ecosystems/orchestrator/orchestrator.go
thomasschafer: On the second point: updated

pkg/ecosystems/legacy/identity.go
```diff
@@ -1,38 +0,0 @@
-package legacy
-
-import (
-	"path"
-	"strings"
-)
-
-// getProjectTargetFileBasedOnType determines if a plugin sets targetFile based on package manager type.
-func getProjectTargetFileBasedOnType(projectType, file string, isWorkspace bool) *string {
```
thomasschafer, line 9: See [here](https://github.com/snyk/cli-extension-dep-graph/pull/166/changes#r3280849063) for reasoning behind deletion

mcombuechen (approved):

james-snyk (approved):

PR #170, opened 2026-05-12
feat: [CMPA-543] Adjusts gradle resolver output to be unpruned DAG

calhar-snyk:

### Summary

Adjust Gradle resolver to produce proper DAG structure and distinguish constraint edges from real dependencies. Two material changes:

1. **Stop creating pruned nodes for visited dependencies** - instead connect to existing nodes to produce a true DAG where nodes can have multiple parents
2. **Add constraint node support** - distinguish constraint edges (from platform BOMs, dependency locking, `constraints {}` blocks) from real artifact dependencies by creating labeled `:constraint` nodes

#### Pruning

The previous behavior created artificial "pruned" leaf nodes for already-visited dependencies, producing tree structures with duplicated pruned nodes rather than true DAGs.

Cycle detection is preserved - cyclic references still create pruned nodes to maintain DAG properties.

#### Constraints

Gradle's dependency resolution includes constraint edges alongside real artifact dependencies, but these serve fundamentally different purposes:

- **Real edges**: Represent actual artifact dependencies that appear in classpaths
- **Constraint edges**: Influence version selection without adding artifacts (from platform BOMs, dependency locking, explicit constraints)

Previously, both were treated identically, causing confusing graph structures where:
- Lock file constraints appeared as direct dependencies of the root project
- Platform BOM constraints appeared as children of BOM components
- No way to distinguish version-influencing constraints from real dependencies

**New behavior**: Constraint edges are detected in the init script and rendered as separate `:constraint`-suffixed nodes with appropriate labels. This:
- Preserves diagnostic information about what constraints influenced resolution
- Allows consumers to distinguish constraints from real dependencies
- Prevents constraint edges from interfering with real dependency graph expansion
- Maintains cleaner separation between version selection and actual artifact relationships

Added integration test fixtures for `constraints {}` DSL, `resolutionStrategy.force`, and `dependencySubstitution` to ensure proper constraint vs. real dependency classification.

### DIFF
--- a/pkg/ecosystems/testdata/fixtures/gradle/with-lock-file/expected_plugin.json
+++ b/pkg/ecosystems/testdata/fixtures/gradle/with-lock-file/expected_plugin.json
@@ -81,47 +81,31 @@
                 "nodeId": "org.codehaus.groovy:groovy@3.0.3"
               },
               {
-                "nodeId": "org.codehaus.groovy:groovy@3.0.3:pruned"
-              },
-              {
                 "nodeId": "com.google.guava:guava@32.0.1-jre"
               },
               {
-                "nodeId": "com.google.guava:guava@32.0.1-jre:pruned"
+                "nodeId": "com.google.guava:failureaccess@1.0.1"
               },
               {
-                "nodeId": "com.google.guava:failureaccess@1.0.1:pruned"
+                "nodeId": "com.google.guava:listenablefuture@9999.0-empty-to-avoid-conflict-with-guava"
               },
               {
-                "nodeId": "com.google.guava:listenablefuture@9999.0-empty-to-avoid-conflict-with-guava:pruned"
+                "nodeId": "com.google.code.findbugs:jsr305@3.0.2"
               },
               {
-                "nodeId": "com.google.code.findbugs:jsr305@3.0.2:pruned"
+                "nodeId": "org.checkerframework:checker-qual@3.33.0"
               },
               {
-                "nodeId": "org.checkerframework:checker-qual@3.33.0:pruned"
+                "nodeId": "com.google.errorprone:error_prone_annotations@2.18.0"
               },
               {
-                "nodeId": "com.google.errorprone:error_prone_annotations@2.18.0:pruned"
-              },
-              {
-                "nodeId": "com.google.j2objc:j2objc-annotations@2.8:pruned"
+                "nodeId": "com.google.j2objc:j2objc-annotations@2.8"
               }
             ]
           },
           {
             "nodeId": "org.codehaus.groovy:groovy@3.0.3",
             "pkgId": "org.codehaus.groovy:groovy@3.0.3",
-            "deps": []
-          },
-          {
-            "nodeId": "org.codehaus.groovy:groovy@3.0.3:pruned",
-            "pkgId": "org.codehaus.groovy:groovy@3.0.3",
-            "info": {
-              "labels": {
-                "pruned": "true"
-              }
-            },
             "deps": []
           },
           {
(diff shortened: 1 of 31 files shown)


pkg/ecosystems/testdata/fixtures/gradle/with-lock-file/expected_plugin.json
```diff
@@ -80,32 +80,26 @@
               {
                 "nodeId": "org.codehaus.groovy:groovy@3.0.3"
               },
-              {
```
calhar-snyk, line 83: This is a moderately weird case. Gradle represents lock file constraints as constraint dependencies on a particular version -- these appear in gradles dependency graph e.g.:
```
+--- org.codehaus.groovy:groovy:[3.0.0,3.0.4) -> 3.0.3
+--- com.google.guava:guava:+ -> 32.0.1-jre
|    +--- com.google.guava:failureaccess:1.0.1
|    +--- com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava
|    +--- com.google.code.findbugs:jsr305:3.0.2
|    +--- org.checkerframework:checker-qual:3.33.0
|    +--- com.google.errorprone:error_prone_annotations:2.18.0
|    \--- com.google.j2objc:j2objc-annotations:2.8
+--- com.google.guava:guava:{strictly 32.0.1-jre} -> 32.0.1-jre (c)
+--- org.codehaus.groovy:groovy:{strictly 3.0.3} -> 3.0.3 (c)
+--- com.google.guava:failureaccess:{strictly 1.0.1} -> 1.0.1 (c)
+--- com.google.guava:listenablefuture:{strictly 9999.0-empty-to-avoid-conflict-with-guava} -> 9999.0-empty-to-avoid-conflict-with-guava (c)
+--- com.google.code.findbugs:jsr305:{strictly 3.0.2} -> 3.0.2 (c)
+--- org.checkerframework:checker-qual:{strictly 3.33.0} -> 3.33.0 (c)
+--- com.google.errorprone:error_prone_annotations:{strictly 2.18.0} -> 2.18.0 (c)
\--- com.google.j2objc:j2objc-annotations:{strictly 2.8} -> 2.8 (c)
```

Having our dep graph representation then have the full tree for each of these is... weird. Particularly for a more complex lock file that perhaps has `A->B->C->D`. All of which would get hoisted and our resulting dep graph would effectively look like:
```
proj
+- A
  \- B
    \- C
      \- D
+- B
  \- C
    \- D
+- C
  \- D
\- D
```

Perhaps the argument then becomes, don't include constraint nodes in the produced dep graph, but that loses useful information.

calhar-snyk: I've adjusted our behaviour.
The issue was that constraint nodes were being treated as real dependency nodes including child dependencies.
e.g.:
```
proj
+- someBom@1.0.0 \\ the including BOM
  +- foo@2.0.0 \\ constraint on foo
    \- bar@1.2.3 \\ child of foo, but not _real_
  \- bar@1.2.3 \\ constraint on bar
\- foo@2.0.0 \\ actual dependency on foo
  \- bar@1.2.3 \\ actual transitive dependency on bar
```

Instead, they are now clearly indicated as a constraint node and we don't recurse through them when generating the intermediate dependency tree. A constraint node is effectively a leaf node that tries to pin the version of some dependency.

We are instead representing the above as:
```
proj
+- someBom@1.0.0 \\ the including BOM
  +- foo@2.0.0 \\ constraint on foo
  \- bar@1.2.3 \\ constraint on bar
\- foo@2.0.0 \\ actual dependency on foo
  \- bar@1.2.3 \\ actual transitive dependency on bar
```

Which accurately matches the output of `gradle dependencies`

calhar-snyk: This also fixes the case with constraints introduced by lockfiles.

Gradle is handling lock file constraints by having them as constraint nodes forcing a strict version choice with the project as their parent. Effectively:
```
proj
+- A@1.2.3 \\ real dependency A
  ...
+- B@2.3.4 \\ real dependency B
  ...
+- A@1.2.3 \\ constraint
+- transitiveA@1.2.3 \\ constraint on some transitive of A
+- B@2.3.4 \\ constraint
+- transitiveB@2.3.4 \\ constraint on some transitive of B
```

This hoisting, combined with is not correctly handling constraint nodes caused us to hoist a reference to every subtree of the _real_ dependency tree. Exploding the number of edges.

tonidevine1 (approved):
This looks good to me! Great explanations in the PR description & comments and thorough testing 👍

PR #171, opened 2026-05-12
feat: optimise target-file scans [CMPA-546]

tonidevine1:
### What this does
This PR scopes Gradle scans to the target sub-project when `--target-file` is set. Previously, when `--target-file` pointed to a sub-project build file, Gradle resolved dependencies for all sub-projects and Go filtered the results afterward. This change passes the target build file path to the init script as -PsnykTargetBuildFile, so only the matching sub-project's dependency resolution task runs. All other sub-projects are skipped entirely.

### DIFF
--- a/pkg/ecosystems/gradle/plugin.go
+++ b/pkg/ecosystems/gradle/plugin.go
@@ -432,7 +432,16 @@ func buildExtraArgs(projectDir string, options *ecosystems.SCAPluginOptions) []s
 		args = append(args, "--init-script", initPath)
 	}
 
-	// Reserved for future flag forwarding (e.g. --configuration, -P flags).
+	// When --target-file is set, scope the Gradle scan to the matching sub-project only.
+	// The init script reads this property and skips dependency resolution for all other projects.
+	if tf := options.Global.TargetFile; tf != nil {
+		absTarget := *tf
+		if !filepath.IsAbs(absTarget) {
+			absTarget = filepath.Join(projectDir, absTarget)
+		}
+		args = append(args, "-PsnykTargetBuildFile="+absTarget)
+	}
+
 	return args
 }
 
--- a/pkg/ecosystems/gradle/init/snyk-deps-init.gradle
+++ b/pkg/ecosystems/gradle/init/snyk-deps-init.gradle
@@ -5,6 +5,10 @@
 //   Line 1:    metadata JSON object
 //   Lines 2+:  one project JSON object per line (newline-delimited JSON)
 //
+// Optional property: -PsnykTargetBuildFile=<absolutePath>
+//   When set, only the project whose build file matches the path runs (skips all others).
+//   Used by the Go plugin when --target-file points to a specific sub-project build file.
+//
 // Uses Gradle's ResolutionResult API (Gradle 4.4+) to programmatically
 // walk the resolved dependency graph per-project, per-configuration.
 //
@@ -15,9 +19,15 @@
 gradle.projectsEvaluated { g ->
     def root = g.rootProject
 
+    // When snykTargetBuildFile is set, only the matching project runs its task.
+    // Otherwise all projects participate (default behaviour).
+    def targetBuildFile = root.findProperty('snykTargetBuildFile')?.toString()
+
     // One task per project, each writing to its own output file.
     // Isolation makes parallel execution safe; snykDependencyGraph depends on all of these.
-    def perProjectTasks = root.allprojects.collect { proj ->
+    def perProjectTasks = root.allprojects
+        .findAll { proj -> !targetBuildFile || getBuildFilePath(proj) == targetBuildFile }
+        .collect { proj ->
         // Resolved at configuration time so outputs.file() can register it for up-to-date checking.
         def outputFile = resolveReportsFile(proj, "snyk-project.ndjson")
 
@@ -228,13 +238,11 @@ def getBuildFilePath(project) {
     try {
         if (project.hasProperty('layout') && project.layout.hasProperty('projectDirectory')) {
             def projectDir = project.layout.projectDirectory.asFile
-            ['build.gradle', 'build.gradle.kts'].each { fileName ->
-                def buildFile = new File(projectDir, fileName)
-                if (buildFile.exists()) {
-                    return buildFile.absolutePath
-                }
+            def found = ['build.gradle', 'build.gradle.kts'].find { fileName ->
+                new File(projectDir, fileName).exists()
             }
-            return new File(projectDir, 'build.gradle').absolutePath
+            return found ? new File(projectDir, found).absolutePath
+                         : new File(projectDir, 'build.gradle').absolutePath
         }
     } catch (Exception e) {
         // Fallback for older Gradle versions or API failures
--- a/pkg/ecosystems/gradle/plugin_test.go
+++ b/pkg/ecosystems/gradle/plugin_test.go
@@ -386,6 +386,32 @@ func TestBuildExtraArgs(t *testing.T) {
 		args := buildExtraArgs(dir, opts)
 		assert.Equal(t, []string{"--init-script", validScript}, args)
 	})
+
+	t.Run("adds snykTargetBuildFile for relative target file", func(t *testing.T) {
+		tf := "app/build.gradle"
+		opts := &ecosystems.SCAPluginOptions{
+			Global: ecosystems.GlobalOptions{TargetFile: &tf},
+		}
+		args := buildExtraArgs("/project", opts)
+		assert.Contains(t, args, "-PsnykTargetBuildFile=/project/app/build.gradle")
+	})
+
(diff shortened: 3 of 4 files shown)


pkg/ecosystems/gradle/plugin.go
```diff
@@ -221,7 +221,21 @@ func (p Plugin) processGradleFile(
 		logger.Attr("init_script", initScriptPath),
 		logger.Attr("gradle_binary", gradleBinary))
 
-	reader, err := runInitScript(ctx, projectDir, gradleBinary, initScriptPath, extraArgs)
+	// When a specific build file is targeted, pass its absolute path to the init script
+	// so only the matching sub-project resolves dependencies, skipping all others.
+	runArgs := extraArgs
+	if options.Global.TargetFile != nil && !isSettingsFile {
+		absTarget := *options.Global.TargetFile
```
calhar-snyk: This logic can probably get pushed to `buildExtraArgs`

calhar-snyk: With whatever unit testing is appropriate

pkg/ecosystems/gradle/plugin.go
```diff
@@ -432,7 +432,16 @@ func buildExtraArgs(projectDir string, options *ecosystems.SCAPluginOptions) []s
 		args = append(args, "--init-script", initPath)
 	}
 
-	// Reserved for future flag forwarding (e.g. --configuration, -P flags).
+	// When --target-file is set, scope the Gradle scan to the matching sub-project only.
+	// The init script reads this property and skips dependency resolution for all other projects.
+	if tf := options.Global.TargetFile; tf != nil {
+		absTarget := *tf
+		if !filepath.IsAbs(absTarget) {
```
tonidevine1, line 439: I don't believe the AI bot's concern is valid for the current code. 

`--target-file` and multi-project discovery are mutually exclusive in discoverGradleFiles, so buildExtraArgs is always called once and used for exactly one file. The root directory is the correct place for resolving the relative target file, and the absolute path we pass to Gradle always matches what getBuildFilePath returns.

pkg/ecosystems/gradle/plugin.go
```diff
@@ -432,7 +432,16 @@ func buildExtraArgs(projectDir string, options *ecosystems.SCAPluginOptions) []s
...
+			absTarget = filepath.Join(projectDir, absTarget)
+		}
+		args = append(args, "-PsnykTargetBuildFile="+absTarget)
```
calhar-snyk, line 442: There's something not right about what we're doing here:
```
❯ ./gradlew --init-script=/Users/calumharrison/code/cli-extension-dep-graph/pkg/ecosystems/gradle/init/snyk-deps-init.gradle snykDependencyGraph -PsnykTargetBuildFile=app/build.gradle.kts
Parallel Configuration Cache is an incubating feature.
Calculating task graph as no cached configuration is available for tasks: snykDependencyGraph
Type-safe project accessors is an incubating feature.

> Task :snykDependencyGraph
SNYK_DEPS_NDJSON /Users/calumharrison/code/playground/nowinandroid/build/reports/snyk-dependency-graph.ndjson

1 problem was found storing the configuration cache.
- Task `:snykDependencyGraph` of type `org.gradle.api.DefaultTask`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/9.4.0/userguide/configuration_cache_requirements.html#config_cache:requirements:disallowed_types

See the complete report at file:///Users/calumharrison/code/playground/nowinandroid/build/reports/configuration-cache/95svowo6j49z3qmce8w5rw8ry/f118j43davyidxgxbgpq40h1q/configuration-cache-report.html

[Incubating] Problems report is available at: file:///Users/calumharrison/code/playground/nowinandroid/build/reports/problems/problems-report.html

BUILD SUCCESSFUL in 6s
5 actionable tasks: 1 executed, 4 up-to-date
Configuration cache entry discarded with 1 problem.
❯ less build/reports/snyk-dependency-graph.ndjson
❯ cat build/reports/snyk-dependency-graph.ndjson
{"gradleVersion":"9.4.0","javaVersion":"25.0.3","javaVendor":"Eclipse Adoptium","javaHome":"/Users/calumharrison/.sdkman/candidates/java/25.0.3-tem","os":"Mac OS X aarch64","generatedAt":"2026-05-13T08:43:37Z","rootProject":{"name":"nowinandroid","group":"","version":"unspecified","path":"/Users/calumharrison/code/playground/nowinandroid"},"maxMemoryMB":4096}
```

calhar-snyk: Hmm, without the flag we still get things like:
```
- Task `:core:common:snykDependencyGraphProject` of type `org.gradle.api.DefaultTask`: cannot serialize object of type 'org.gradle.api.internal.project.LifecycleAwareProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/9.4.0/userguide/configuration_cache_requirements.html#config_cache:requirements:disallowed_types
- Task `:core:data-test:snykDependencyGraphProject` of type `org.gradle.api.DefaultTask`: cannot serialize object of type 'org.gradle.api.internal.project.LifecycleAwareProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
```

But we still get the correct number of projecs I think:
```
❯ wc -l build/reports/snyk-dependency-graph.ndjson
      46 build/reports/snyk-dependency-graph.ndjson
```

tonidevine1: Yeah so I believe the bug was in `getBuildFilePath`. The return inside `.each` didn't return from the function in Groovy , it just skipped to the next iteration. So `getBuildFilePath `always fell through to return `build.gradle`, even for KTS-only projects. I've fixed this by switching to `.Find`. 

On the second point that would be expected. Without the flag all projects participate as before, so the full scan was unaffected.

calhar-snyk (approved):

tonidevine1:
> ## PR Reviewer Guide 🔍
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Incorrect Path Resolution**](https://github.com/snyk/cli-extension-dep-graph/pull/171/changes#diff-9e410e6b4882afc044abf7ec92d5c0697078c8012c42fd1b1bc8b4c727248b36R439-R441) 🟠 [major]
> The logic in `buildExtraArgs` resolves a relative `TargetFile` against the `projectDir` passed as an argument. However, `processGradleFiles` calls `buildExtraArgs` using the root scan directory (`dir`). If a user executes the scan from a different working directory and provides a path relative to their current terminal location, `filepath.Join(projectDir, absTarget)` will produce a non-existent path. The `TargetFile` should ideally be resolved against the initial execution directory before being passed into the plugin's internal resolution logic.
> 
> ```go
> if !filepath.IsAbs(absTarget) {
> 	absTarget = filepath.Join(projectDir, absTarget)
> }
> ```
> 
> 📚 **Repository Context Analyzed**
> 
> **This review considered 8 relevant code sections from 5 files (average relevance: 0.81)**
> 
> 
> * [pkg/ecosystems/gradle/executor.go](https://github.com/snyk/cli-extension-dep-graph/blob/783385a6003369af3e26ce9bfba9b047c1e05ea6/pkg/ecosystems/gradle/executor.go/#L1-L125) (lines 1-125)
> * [pkg/ecosystems/options.go](https://github.com/snyk/cli-extension-dep-graph/blob/783385a6003369af3e26ce9bfba9b047c1e05ea6/pkg/ecosystems/options.go/#L1-L177) (lines 1-177)
> * [pkg/ecosystems/gradle/README.md](https://github.com/snyk/cli-extension-dep-graph/blob/783385a6003369af3e26ce9bfba9b047c1e05ea6/pkg/ecosystems/gradle/README.md/#L96-L167) (2 segments, lines 96-167)
> * [pkg/ecosystems/gradle/validation.go](https://github.com/snyk/cli-extension-dep-graph/blob/783385a6003369af3e26ce9bfba9b047c1e05ea6/pkg/ecosystems/gradle/validation.go/#L1-L64) (lines 1-64)
> * [pkg/ecosystems/gradle/plugin.go](https://github.com/snyk/cli-extension-dep-graph/blob/783385a6003369af3e26ce9bfba9b047c1e05ea6/pkg/ecosystems/gradle/plugin.go/#L194-L418) (3 segments, lines 194-418)

The paths from discovery are absolute. FindFiles calls `filepath.Abs` on `rootDir` before returning results, and `FindResult.Path` is explicitly documented as an absolute path. By the time `buildExtraArgs` is called, `dir` is already absolute, so joining a relative `TargetFile` against it produces a correct absolute path for `-PsnykTargetBuildFile`. The same join logic is also used in `convertProjects` for target-file filtering, so the behaviour is consistent throughout the plugin. No issue here.

PR #172, opened 2026-05-13
feat: [CMPA-540] Adds configuration-attribute flag handling to gradle resolver

calhar-snyk:

### Summary
Implements attribute-based configuration filtering for feature parity with the legacy `snyk-gradle-plugin`. The `--configuration-attributes` flag accepts comma-separated `key:value` pairs (e.g., `"buildtype:release,usage:java-runtime"`) to filter Gradle configurations by their declared attributes.
This feature is particularly valuable for Android builds where filtering by attributes like `buildtype:release` or `usage:java-runtime` can significantly reduce scan time by avoiding resolution of unwanted configuration variants.
#### Core Implementation
- **Added `ConfigurationAttributes` field** to `GradleOptions` with validation
- **Enhanced `buildExtraArgs`** to forward attributes as `-PsnykConfAttr=<spec>` property to Gradle
- **Updated `snyk-deps-init.gradle`** with attribute matching logic supporting:
  - Exact matches on logical attribute names (`org.gradle.usage`)
  - Last segment matching (`usage` matches `org.gradle.usage`) 
  - Type simple class name matching (`buildtype` matches `BuildTypeAttr`)
  - Legacy substring fallback for backward compatibility
#### Performance Optimizations
- **Parse attribute specification once upfront** rather than per-project
- **Early filtering in Gradle script** avoids resolving unwanted configurations


pkg/ecosystems/gradle/plugin_integration_test.go
```diff
@@ -173,6 +173,16 @@ func pluginTestCases() map[string]PluginTestCase {
 			Options:      ecosystems.NewPluginOptions().WithGradleConfigurationMatching("^runtimeClasspath$"),
 			ExpectedFile: "expected_plugin_exact_runtime.json",
 		},
+		// Exercises --configuration-attributes flag functionality using a fixture with dependencies
+		// in multiple configurations. Standard Java plugin sets attributes like org.gradle.usage
+		// on configurations. This validates attribute-based filtering works correctly.
+
+		// Filter by usage:java-runtime - should include runtime configurations
+		"configuration_attributes_java_runtime": {
```
tonidevine1, line 182: I think this is where the CI failure is coming from. On Gradle 6 there are legacy configurations with no usage attribute at all, so they pass this filter unconditionally and bring commons-lang3 (a compileOnly dep) along with them.

Would it make more sense to exclude configs that don't have the attribute you're filtering on?

calhar-snyk: Plausibly -- I'll dig into the exact nature of the failure

calhar-snyk: Yeah, it's exactly that. Things like:
```
DEBUG: Config testCompileOnly has 0 attributes: [:]
DEBUG: Config testCompileOnly ACCEPTED - all constraints satisfied
DEBUG: Config testRuntime has 0 attributes: [:]
DEBUG: Config testRuntime ACCEPTED - all constraints satisfied
```
I should have noticed this was a potential problem when doing the port -- will check what behaviour we actually want to have here. Because, although the logic here is consistent with the legacy plugin, I think it's wrong.

calhar-snyk: Adjusted the logic to filter any configurations that don't explicitly match the provided `configuration-attributes`

tonidevine1 (approved):

PR #176, opened 2026-05-15
feat: [CMPA-542] Adds support in Gradle resolver for `--include-provenance`

calhar-snyk:

### Summary

Adds support for the `--include-provenance` global flag to the Gradle resolver, enabling SHA1 checksum collection and PURL generation with checksum qualifiers.

##### Checksum Collection Strategy
- **Leverages Gradle's internal `ChecksumService`** for optimal performance - uses Gradle's cached checksums when available (Gradle 6+)
- **Falls back to manual SHA1 calculation** when ChecksumService unavailable
- **O(1) artifact lookup** - builds a `ComponentIdentifier → List<ResolvedArtifactResult>` map per configuration to avoid O(N²) artifact searches

##### PURL Generation
- **Conditional PURL generation** - only generates PURLs when `--include-provenance` is enabled
- **Checksum qualifiers** - adds `?checksum=sha1:abc123...` to PURLs when checksums are available
- **Base PURLs for missing checksums** - generates `pkg:maven/group/artifact@version` even when checksum unavailable

##### BOM/Platform Dependency Handling
- **Skips provenance for BOMs** - BOM/platform dependencies (like `spring-boot-dependencies`) don't have physical artifacts to checksum, so provenance collection is skipped for these metadata-only dependencies

##### Init Script Enhancements
- **Provenance threading** - passes `includeProvenance` flag through the resolution pipeline
- **Artifact type detection** - extracts file extension as artifact type for dependency classification
- **Error resilience** - checksum failures don't break dependency graph generation

The implementation aligns with Maven plugin behaviour, focusing provenance collection on actual downloadable artifacts rather than metadata dependencies.

### DIFF
--- a/pkg/ecosystems/gradle/init/snyk-deps-init.gradle
+++ b/pkg/ecosystems/gradle/init/snyk-deps-init.gradle
@@ -23,6 +23,9 @@ gradle.projectsEvaluated { g ->
     // Otherwise all projects participate (default behaviour).
     def targetBuildFile = root.findProperty('snykTargetBuildFile')?.toString()
 
+    // Check if provenance (checksums) should be included
+    def includeProvenance = root.findProperty('snykIncludeProvenance')?.toBoolean() ?: false
+
     // One task per project, each writing to its own output file.
     // Isolation makes parallel execution safe; snykDependencyGraph depends on all of these.
     def perProjectTasks = root.allprojects
@@ -42,7 +45,7 @@ gradle.projectsEvaluated { g ->
             }
             doLast {
                 outputFile.parentFile.mkdirs()
-                outputFile.text = groovy.json.JsonOutput.toJson(buildProjectEntry(proj))
+                outputFile.text = groovy.json.JsonOutput.toJson(buildProjectEntry(proj, includeProvenance))
             }
         }
     }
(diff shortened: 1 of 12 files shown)


pkg/ecosystems/gradle/init/snyk-deps-init.gradle
```diff
@@ -218,15 +226,61 @@ def walkDependencies(component, Set<String> processed, Set<String> ancestry) {
...
+
+        // Try different API approaches for Gradle version compatibility
+        // We use capability-based detection rather than gradle.gradleVersion parsing
+        // to handle custom builds, snapshots, and vendor distributions robustly
+        try {
+            // Gradle 7+ approach - resolvedArtifacts property
+            def artifactsProvider = artifactCollection.resolvedArtifacts
+            resolvedArtifacts = artifactsProvider.get()
+        } catch (Exception e1) {
+            // Gradle 6 approach - direct iteration over artifact collection
```
tonidevine1, line 247: Just wondered about support for Gradle 5.0, does anything change for this? If not maybe we should update the comment so that we explicitly mention support for Gradle 5.0+

tonidevine1: We discussed in the call that we are supporting Gradle 6+ so this comment is resolved.

tonidevine1 (approved):

PR #177, opened 2026-05-15
feat: add support for sub project legacy flag [CMPA-548]

tonidevine1:

### What this does

To keep things consistent with the legacy plugin we are adding the `--sub-project` flag. 

https://snyksec.atlassian.net/jira/software/c/projects/CMPA/boards/5466?assignee=60162bfac8c36c00698fa2d4&selectedIssue=CMPA-548

### DIFF
--- a/pkg/ecosystems/options.go
+++ b/pkg/ecosystems/options.go
@@ -53,7 +53,8 @@ type GradleOptions struct {
 	// ConfigurationAttributes filters configurations by attribute values (key:value,key:value).
 	ConfigurationAttributes string `arg:"--configuration-attributes"`
 	// SubProject restricts scanning to a single named Gradle sub-project.
-	SubProject string `arg:"--gradle-sub-project"`
+	// Accepts both --gradle-sub-project and --sub-project (legacy alias).
+	SubProject string `arg:"--gradle-sub-project,--sub-project"`
 	// AllSubProjects scans all sub-projects in a multi-project build.
 	AllSubProjects bool `arg:"--all-sub-projects"`
 	// InitScript overrides the built-in init script with a user-supplied path.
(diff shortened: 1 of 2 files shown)


pkg/ecosystems/options.go
```diff
@@ -53,7 +53,8 @@ type GradleOptions struct {
 	// ConfigurationAttributes filters configurations by attribute values (key:value,key:value).
 	ConfigurationAttributes string `arg:"--configuration-attributes"`
 	// SubProject restricts scanning to a single named Gradle sub-project.
-	SubProject string `arg:"--gradle-sub-project"`
+	// Accepts both --gradle-sub-project and --sub-project (legacy alias).
+	SubProject string `arg:"--gradle-sub-project,--sub-project"`
```
calhar-snyk, line 57: Silly hypothetical: What happens if someone does `--gradle-sub-project=foo --sub-project=bar`? Is it last one wins?

tonidevine1: oh that's a really interesting point

tonidevine1: Yes this is a case of last one wins. I guess does this even matter if both flags have identical behaviour?

tonidevine1: In the legacy TypeScript CLI, --gradle-sub-project gets aliased to --sub-project before the plugin is called (see [here](https://github.com/snyk/cli/blob/main/src/cli/args.ts#L272)), so the legacy plugin only ever sees one unified subProject field.


So this would be the same for the legacy flow also?

calhar-snyk: Looks less like last one wins and more `gradle-sub-project` wins, since it's the one that we splat over the top:
```
  const aliases = {
    gradleSubProject: 'subProject',
    container: 'docker',
  };
  for (const argAlias in aliases) {
    if (argv[argAlias]) {
      const target = aliases[argAlias];
      argv[target] = argv[argAlias];
      delete argv[argAlias];
    }
  }
```

I don't _think_ we care that much about picture perfect accuracy here. It's a weird edge case that shouldn't be showing up anyway. Priority is preserving operation of the individual flags

calhar-snyk (approved):

PR #179, opened 2026-05-19
feat: consolidate print-graph flags and add error-only entry checking

neema-beglou-snyk:
## What does this PR do?

Adds support for the new `FlagPrune` flag in the dep-graph router, which is the central coordination point between the CLI (producer) and all consumers (SBOM, snyk-delta, container-cli).

### Flag routing

`ChooseGraphArgument()` now checks `FlagPrune` first (takes priority over legacy flags):

| Config | CLI args produced | Output parser |
|---|---|---|
| `FlagPrune=true` | `--print-graph --prune` | JSONL |
| `FlagPrune=false` | `--print-graph --jsonl` | JSONL |
| Legacy flags (no FlagPrune) | unchanged | unchanged |

### Error-only entry bridge

`HandleLegacyResolution` now scans JSONL for error-only entries (entries with `error` but no `depGraph`). Behavior is controlled by `shouldThrowOnErrors(config)`:

- **Returns `true` (default):** throws the error as `(nil, snyk_errors.Error)` — preserves the contract for container-cli and snyk-delta which expect thrown errors.
- **Returns `false`:** passes error-only entries through as `workflow.Data` with embedded errors — used by SBOM (via `fail-fast=false`) which reads them via `GetErrorList()`.

`shouldThrowOnErrors` returns false when:
- Legacy `-with-errors` flags are set (backward compat)
- `fail-fast=false` is set (SBOM with `--allow-incomplete-sbom`)

### SBOM resolution simplified

`executeLegacyWorkflow` in `sbom_resolution.go` now unsets all legacy flags and sets `FlagPrune=false` + `FlagFailFast=false`. This replaces the previous conditional logic around `print-output-jsonl-with-errors`.

### Error chain preservation

`ExtractLegacyCLIError` now wraps both the parsed error-catalog error and the original `exec.ExitError` using `fmt.Errorf("%w: %w", ...)`, so exit codes are preserved through the error chain.

### External plugin updated

`plugin.go` switches from `--print-effective-graph-with-errors` to `--print-graph --prune`.

## Where should the reviewer start?

`internal/legacycli/legacy_resolution.go` — the `shouldThrowOnErrors` function and the error-checking block in `HandleLegacyResolution`. Then `ChooseGraphArgument` and `PrepareLegacyFlags` for the flag routing.

## Risk assessment: Medium

- Legacy flag paths are preserved — no existing consumers break.
- `FlagPrune` takes strict priority over legacy flags when both are present.
- SBOM path is simplified but functionally equivalent (same JSONL output, same error embedding).
- Main risk is the `shouldThrowOnErrors` bridge — if a consumer doesn't set the right config signals, it could get thrown errors when it expects embedded ones (or vice versa). Mitigated by tests covering all signal combinations.

## Test coverage

- 3 tests for `HandleLegacyResolution` error checking (throw vs embed, error-only entries)
- 7 tests for `shouldThrowOnErrors` (all flag combinations)
- 3 tests for `ChooseGraphArgument` with `FlagPrune` (true, false, priority over legacy)
- 1 test for `PrepareLegacyFlags` suppressing legacy flags when `FlagPrune` is set
- 2 tests for `executeLegacyWorkflow` with new `FlagPrune`/`FlagFailFast` config
- 1 test for `ExtractLegacyCLIError` preserving `exec.ExitError` in the chain
- New test fixture: `jsonl_error_only_output`

## Relevant tickets

- CSENG-190 (CLI)
- Consumers of extension-dep-graph: CSENG-192 (SBOM), CSENG-193 (snyk-delta), CSENG-198 (container-cli)


internal/legacycli/legacy_resolution.go
```diff
@@ -269,5 +288,32 @@ func PrepareLegacyFlags(argument string, cfg configuration.Configuration, logger
...
+// error-only JSONL entries. Returns false when a consumer has opted in to
+// handling embedded errors itself (legacy -with-errors flags, or fail-fast=false).
+// DEPRECATION: Remove this function along with the throw-on-error block in
+// HandleLegacyResolution once all consumers handle embedded errors directly.
+func shouldThrowOnErrors(config configuration.Configuration) bool {
+	if config.GetBool(workflow.FlagPrintEffectiveGraphWithErrors) ||
+		config.GetBool(workflow.FlagPrintOutputJsonlWithErrors) {
+		return false
+	}
+	if config.IsSet(workflow.FlagFailFast) && !config.GetBool(workflow.FlagFailFast) {
```
thomasschafer, line 315: This seems like some logic that might be quite hard to debug in future, i.e. `--fail-fast` being `false` behaving differently to the `--fail-fast` flag not being set at all. What do you think, e.g. we could add an explicit flag to include errors, or something similar? Let me know if I'm missing anything

mcombuechen (approved):

PR #181, opened 2026-05-20
refactor: use variadic options for plugin registration

mcombuechen:
### What this does

Changes the registry's `register` function signature to use more Go-idiomatic, variadic options.

### DIFF
--- a/pkg/ecosystems/orchestrator/opts.go
+++ b/pkg/ecosystems/orchestrator/opts.go
@@ -1,0 +1,18 @@
+package orchestrator
+
+type registerOpt func(*PluginRegistry, *pluginEntry)
+
+func withFeatureFlag(flag flag) registerOpt {
+	return func(reg *PluginRegistry, entry *pluginEntry) {
+		if reg.ictx.GetConfiguration().GetBool(flag.Key) {
+			return
+		}
+		entry.skip = true
+	}
+}
+
+func withPluginDependencies(deps ...string) registerOpt {
+	return func(_ *PluginRegistry, entry *pluginEntry) {
+		entry.dependencies = deps
+	}
+}
--- a/pkg/ecosystems/orchestrator/registry_test.go
+++ b/pkg/ecosystems/orchestrator/registry_test.go
@@ -51,9 +51,9 @@ func TestPluginRegistry_OrderPreservedWithoutDependencies(t *testing.T) {
 		plugins: make([]ecosystems.SCAPlugin, 0),
 	}
 
-	require.NoError(t, r.register(&mockPlugin{name: "plugin-a"}, ""))
-	require.NoError(t, r.register(&mockPlugin{name: "plugin-b"}, ""))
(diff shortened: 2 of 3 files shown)


pkg/ecosystems/orchestrator/opts.go
```diff
@@ -0,0 +1,18 @@
+package orchestrator
+
+type registerOpt func(*PluginRegistry, *pluginEntry)
+
+func withFeatureFlag(flag string) registerOpt {
```
JamesPatrickGill: suggestion/nit: this name could be clearer as this is really a gate.

`requireFeatureFlag` - feels clearer

pkg/ecosystems/orchestrator/opts.go
```diff
@@ -0,0 +1,18 @@
...
+	return func(reg *PluginRegistry, entry *pluginEntry) {
+		if reg.ictx.GetConfiguration().GetBool(flag) {
+			return
+		}
+		entry.skip = true
+	}
+}
+
+func withPluginDependecies(deps ...string) registerOpt {
```
JamesPatrickGill: `withPluginDependecies` should be `withPluginDependencies`

JamesPatrickGill (commented):
Two small comments 🙏

pkg/ecosystems/orchestrator/registry_test.go
```diff
@@ -264,6 +264,25 @@ func TestPluginRegistry_ResolveDepgraphs_ContinuesWhenNoResults(t *testing.T) {
...
+	r.ictx.GetConfiguration().Set("this-ff-is-disabled", false)
+	// should get skipped
+	err := r.register(mockPlugin{name: "plugin-a"}, withFeatureFlag("this-ff-is-disabled"))
+	require.NoError(t, err)
+	// should get registered
+	err = r.register(mockPlugin{name: "plugin-b"})
+	require.NoError(t, err)
+
+	assert.Len(t, r.entries, 1, "registry has 1 entry")
+	assert.Len(t, r.plugins, 1, "registry has 1 plugin")
```
cjheppell: This test would still pass if we messed up the logic as we're only asserting on the number of entries, not the specific plugins that were registered.

In other words - if we inadvertently registered `plugin-a` and not `plugin-b` (because our logic was incorrect), this test wouldn't catch that.

pkg/ecosystems/orchestrator/opts.go
mcombuechen: I wanted to keep this in line with [Go's functional options pattern](https://go-patterns.dev/creational/functional-options), where functions commonly start with `With...`. How about `withFeatureFlagCheck`?

pkg/ecosystems/orchestrator/registry_test.go
mcombuechen: Changed it to a more deterministic assertion.

james-snyk (approved):

PR #188, opened 2026-05-22
fix: preserve monitor identity for poetry.lock

thomasschafer:
This PR adds a small fix to `Identity.TargetFile` to ensure it matches the logic in Registry. Currently, `depGraphOutputToSCAResult` forwards the legacy CLI's `targetFileFromPlugin` field into `Identity.TargetFile`, and that value flows downstream into the monitor project identity. If it ever diverges from what the legacy CLI itself sends, existing monitor projects are not found and duplicates get created.

Registry applies one compatibility transform on input: exact-string `"poetry.lock"` → `"pyproject.toml"`. `snyk-python-plugin` already [produces](https://github.com/snyk/snyk-python-plugin/blob/718e100/lib/dependencies/inspect-implementation.ts#L55-L74) `"pyproject.toml"`, so the transform is unused today, but if that invariant ever breaks (e.g. changes to the Poetry resolver) we would no longer be matching the behaviour in Registry.

### DIFF
--- a/pkg/ecosystems/legacy/plugin.go
+++ b/pkg/ecosystems/legacy/plugin.go
@@ -13,6 +13,7 @@ import (
 	gafworkflow "github.com/snyk/go-application-framework/pkg/workflow"
 
 	"github.com/snyk/cli-extension-dep-graph/internal/legacycli"
+	"github.com/snyk/cli-extension-dep-graph/internal/ptrutil"
 	"github.com/snyk/cli-extension-dep-graph/internal/workflow"
 	"github.com/snyk/cli-extension-dep-graph/pkg/depgraph/parsers"
 	"github.com/snyk/cli-extension-dep-graph/pkg/ecosystems"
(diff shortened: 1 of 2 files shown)


pkg/ecosystems/legacy/plugin.go
```diff
@@ -100,16 +101,27 @@ func buildLegacyConfig(src configuration.Configuration, opts *ecosystems.SCAPlug
 	return cfg
 }
 
+// mapToTargetFileIdentity mirrors registry's mapping from `targetFileFromPlugin` to target file identifier.
+func mapToTargetFileIdentity(ctx context.Context, log logger.Logger, targetFile *string) *string {
+	if ptrutil.DerefOr(targetFile, "") == "poetry.lock" {
+		log.Info(ctx, "rewriting targetFileFromPlugin: 'poetry.lock' → 'pyproject.toml'")
+		return new("pyproject.toml")
```
mcombuechen, line 108: Might be worth checking if this breaks consumers downstrem. IIRC `new("some string")` is a fairly recent Go feature.

PR #190, opened 2026-05-22
feat: surface unresolved dependencies [CMPA-553]

tonidevine1:

### What this does

As part of ongoing support work the snyk-gradle-plugin is in the process of being rewritten using modern gradle apis to solve a few support bugs: [feat: Update init.gradle to use non-deprecated Gradle APIs by cjheppell · Pull Request #337 · snyk/snyk-gradle-plugin](https://github.com/snyk/snyk-gradle-plugin/pull/337/changes#diff-c0f728edeefc1feb5762c48b6097c42a81a6ee217d135ed1727ead0c0f04e425)

This PR addresses the issue of surfacing unresolved dependencies. Previously, dependencies that Gradle couldn't resolve were silently dropped from the dep graph. They now appear as labeled leaf nodes so consumers can see what failed. 

[Jira](https://snyksec.atlassian.net/jira/software/c/projects/CMPA/boards/5466?assignee=60162bfac8c36c00698fa2d4&selectedIssue=CMPA-553)

### DIFF
--- a/pkg/ecosystems/gradle/depgraph.go
+++ b/pkg/ecosystems/gradle/depgraph.go
@@ -112,8 +112,20 @@ type depGraphContext struct {
 //     with a `:constraint` node-ID suffix. They do not affect `processed`, so a
 //     real edge to the same component will still be expanded normally.
 func (ctx *depGraphContext) addDep(dep *gradleDep, parentID string, processed map[string]bool) error {
-	if dep.ID == "" || dep.Unresolved {
+	if dep.ID == "" {
 		return nil
+	}
+
+	if dep.Unresolved {
+		// ":unresolved" suffix avoids colliding with a resolved node for the same coordinate in another configuration.
+		nodeID, name, version := depNodeParts(dep.ID)
+		unresolvedID := nodeID + ":unresolved"
+		ctx.builder.AddNode(unresolvedID, createPkgInfo(name, version, nil, ctx.provenanceMap != nil),
+			depgraph.WithNodeInfo(&depgraph.NodeInfo{
+				Labels: map[string]string{"unresolved": "true"},
+			}),
+		)
+		return ctx.connectOnce(parentID, unresolvedID)
 	}
 
 	nodeID, name, version := depNodeParts(dep.ID)
--- a/pkg/ecosystems/gradle/init/snyk-deps-init.gradle
+++ b/pkg/ecosystems/gradle/init/snyk-deps-init.gradle
@@ -158,10 +158,25 @@ def resolveConfiguration(config, includeProvenance = false) {
     def resolutionResult = config.incoming.resolutionResult
     def rootComponent = resolutionResult.root
 
+    // Per-config set of resolved group:module names; used to avoid surfacing false unresolved deps that strict locking emits alongside real resolved deps.
+    def resolvedNames = new HashSet<String>()
+    try {
+        resolutionResult.allComponents.each { component ->
+            def mv = component.moduleVersion
+            if (mv != null && mv.group && mv.name) {
+                // .toString() is load-bearing — GStrings and Strings hash differently.
+                resolvedNames.add("${mv.group}:${mv.name}".toString())
+            }
+        }
+    } catch (Exception ignored) {
+        // allComponents may be unavailable in very old Gradle versions; proceed
+        // without noise filtering rather than failing the whole configuration.
+    }
+
     def rootId = componentId(rootComponent)
     def processed = new HashSet<String>()
     def ancestry = new HashSet<String>()
-    def tree = walkDependencies(rootComponent, processed, ancestry)
+    def tree = walkDependencies(rootComponent, processed, ancestry, resolvedNames)
 
     // Build artifact lookup map once per configuration for O(1) lookups
     def artifactLookup = null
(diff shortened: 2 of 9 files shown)


pkg/ecosystems/testdata/fixtures/gradle/same-name-subprojects/subproj/build.gradle
```diff
@@ -1,7 +1,3 @@
 plugins {
     id 'java'
 }
-
-dependencies {
```
calhar-snyk, line 5: Why are we removing this dependency here?

tonidevine1: This dependency is missing the artifact name so as it cannot be resolved was silently dropped before. Since we are now surfacing unresolved dependencies this was no longer dropped and causing tests to fail which is why I removed it. 

Please let me know if there is a use-case I am overlooking here and I can add this back in appropriately ☺️

calhar-snyk: Ahh, it's just a bad dependency and a typo. Should instead be `com.google.guava:guava:32.0.1-jre` and updating the fixtures as appropriate.
Well, or that or one of the `31` versions, I recalled the gradle 6 test image has issues handling guava@32

pkg/ecosystems/gradle/depgraph.go
```diff
@@ -112,10 +112,28 @@ type depGraphContext struct {
...
 		return nil
 	}
 
+	if dep.Unresolved {
+		// ":unresolved" suffix avoids colliding with a resolved node for the same coordinate in another configuration.
+		nodeID, name, version := depNodeParts(dep.ID)
+		unresolvedID := nodeID + ":unresolved"
+		labels := map[string]string{"unresolved": "true"}
+		if dep.Reason != "" {
+			reason := []rune(dep.Reason)
```
calhar-snyk: Do we really need to manipulate a string to a `[]rune` to take its len and slice it?

tonidevine1: This was the address one of the PR bot's comments but I can remove this to see if it is needed!

pkg/ecosystems/gradle/init/snyk-deps-init.gradle
```diff
@@ -224,14 +239,29 @@ def walkDependencies(component, Set<String> processed, Set<String> ancestry) {
...
+                : (dep.hasProperty('requested') ? dep.requested : null)
+            // Skip project deps and other non-Maven selectors that have no GAV coordinate.
+            if (selector == null || !selector.hasProperty('group') || !selector.hasProperty('module')) return
+            // Skip lock-rejected requests — the module resolved fine at the locked version.
+            def selectorName = "${selector.group}:${selector.module}".toString()
+            if (resolvedNames.contains(selectorName)) return
+            def selectorVersion = 'unspecified'
+            try {
+                def vc = selector.hasProperty('versionConstraint') ? selector.versionConstraint : null
+                selectorVersion = vc?.requiredVersion ?: vc?.preferredVersion ?: selector.version ?: 'unspecified'
```
calhar-snyk: Huh, `x ?:y` is shorthand for `x ? x : y`? neat.

pkg/ecosystems/gradle/init/snyk-deps-init.gradle
```diff
@@ -224,14 +239,29 @@ def walkDependencies(component, Set<String> processed, Set<String> ancestry) {
...
+            } catch (Exception ignored) {
```
calhar-snyk: Is this ever actually catching anything? I wouldn't think so unless `selector` itself is `null`.
Ahh, if `selector` doesn't have a `version` property -- but then our catch is presumably in trouble anyway?

calhar-snyk: I feel like this can be simplified to something like this, which I believe should be null-safe and does aware with the pre-definition of `selectorVersion` and the try/catch
```
def vc = selector.hasProperty('versionConstraint') ? selector.versionConstraint : null
def simpleVersion = selector.hasProperty('version') ? selector.version : 'unspecified'
selectorVersion = vc?.requiredVersion ?: vc?.preferredVersion ?: simpleVersion
```

pkg/ecosystems/gradle/depgraph.go
```diff
@@ -112,10 +112,28 @@ type depGraphContext struct {
...
 
+	if dep.Unresolved {
+		// ":unresolved" suffix avoids colliding with a resolved node for the same coordinate in another configuration.
+		nodeID, name, version := depNodeParts(dep.ID)
+		unresolvedID := nodeID + ":unresolved"
+		labels := map[string]string{"unresolved": "true"}
+		if dep.Reason != "" {
+			reason := dep.Reason
+			if len(reason) > 200 {
+				reason = reason[:200]
```
calhar-snyk: Worth adding a small test for this?

calhar-snyk: That said, I am a bit wary of passing a length restricted gradle error through with sanitising the contents.

tonidevine1: Maybe we don't want gradle text at all. And we just mark the node with an unresolved label?

calhar-snyk: Seems fine to me in the short term. If we decide we want it later, we can add this back

pkg/ecosystems/gradle/depgraph.go
calhar-snyk: I looked it up myself separately. Apparently converting to runes takes it as utf-8, while treating it as a string basically treats each byte individually. Given it's Gradle, unicode safety may actually be relevant there. But the block itself is no longer relevant as we've removed the reason. So it goes.

calhar-snyk (approved):

PR #192, opened 2026-05-27
feat: finding actually resolvable configurations [CMPA-567]

tonidevine1:

### Background

Gradle automatically creates default and archives configurations for every Java project. These are marked as both `canBeResolved` and `canBeConsumed`, with no directly-declared dependencies. They aggregate via `extendsFrom` from the real classpath configurations.

Previously the init script walked all `canBeResolved` configurations, which included these. Walking them via ResolutionResult pulls in inherited transitives that don't belong to any real classpath the project builds against, and can surface deps not covered by a lockfile.

### What this does

We now skip configurations that are both consumable and have no directly-declared dependencies. This change matches the approach taken in the v6 snyk-gradle-plugin branch (see [here](https://github.com/snyk/snyk-gradle-plugin/pull/337)). 

[Jira](https://snyksec.atlassian.net/jira/software/c/projects/CMPA/boards/5466?assignee=60162bfac8c36c00698fa2d4&selectedIssue=CMPA-567)

### DIFF
--- a/pkg/ecosystems/gradle/init/snyk-deps-init.gradle
+++ b/pkg/ecosystems/gradle/init/snyk-deps-init.gradle
@@ -113,12 +113,15 @@ def buildProjectEntry(project, confAttrSpec, includeProvenance = false) {
         if (!config.canBeResolved) {
             return false
         }
-        
+        // Skip auto-created publication configs (default, archives) which have no
+        // directly-declared deps. User-defined configs with declared deps are kept.
+        if (config.canBeConsumed && config.dependencies.isEmpty()) {
+            return false
+        }
         // Apply attribute filter if specified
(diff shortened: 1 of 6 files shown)


pkg/ecosystems/gradle/init/snyk-deps-init.gradle
```diff
@@ -113,12 +113,14 @@ def buildProjectEntry(project, confAttrSpec, includeProvenance = false) {
         if (!config.canBeResolved) {
             return false
         }
-        
-        // Apply attribute filter if specified
```
calhar-snyk, line 117: Nitpick, but we've dropped this comment seemingly for no reason. It's value is arguable though

tonidevine1: oops, I've missed that! Thank you

calhar-snyk (approved):

PR #194, opened 2026-05-28
refactor!: stream SCAResults through SCAPlugin via per-graph callback

cjheppell:
## Summary

To avoid issues in certain ecosystems where dependency graphs might be extremely large, this changes the interface to provide a callback function for when a dep graph is resolved. This allows consumers to determine their own processing pathway - aggregating results in memory if desired, or writing to disk as results arrive to avoid potentially huge memory consumption.

```go
// Before
BuildDepGraphsFromDir(ctx, log, dir, opts) (*PluginResult, error)

// After
BuildDepGraphsFromDir(ctx, log, dir, opts, onGraph func(SCAResult) error) error
```

⚠️ **This is a breaking API change.** All in-repo plugins (bun, bazel, gradle, legacy, pip, pipenv, uv) and consumers (`orchestrator/registry`, `depgraph/sbom_resolution`) are updated in this PR. External consumers will need to update their own consuming code.

## Concurrency contract

The plugin must invoke `onGraph` serially. Plugins that parallelise graph construction internally (pip, pipenv) wrap the call site in a `sync.Mutex` to ensure serial callback execution.

## Review order

Review this commit by commit.

I took the decision to NOT ensure compilation completeness per-commit, and optimised for reviewers readability. This does mean `git bisect` around these commits will not compile, but I think this is an acceptable tradeoff.

### DIFF
--- a/pkg/ecosystems/bazel/plugin.go
+++ b/pkg/ecosystems/bazel/plugin.go
@@ -32,7 +32,8 @@ func (p Plugin) BuildDepGraphsFromDir(
 	log logger.Logger,
 	dir string,
 	options *ecosystems.SCAPluginOptions,
-) (*ecosystems.PluginResult, error) {
+	onGraph func(ecosystems.SCAResult) error,
+) error {
 	if log == nil {
 		log = logger.Nop()
 	}
@@ -41,23 +42,28 @@ func (p Plugin) BuildDepGraphsFromDir(
 	if err != nil {
 		if errors.Is(err, errNoBazelOptionFound) {
 			log.Debug(ctx, "no bazel option found, skipping bazel dependency graph resolution")
-			return &ecosystems.PluginResult{}, nil
+			return nil
 		}
-		return nil, fmt.Errorf("failed to initialize bazel resolver: %w", err)
+		return fmt.Errorf("failed to initialize bazel resolver: %w", err)
 	}
 	log.Debug(ctx, "using bazel resolver", logger.Attr("type", resolver.packageManagerName()))
 
 	targets, err := resolver.findTargets(ctx, options)
 	if err != nil {
-		return nil, fmt.Errorf(errQueryBazelTargetsFmt, err)
+		return fmt.Errorf(errQueryBazelTargetsFmt, err)
 	}
 	log.Debug(ctx, "found bazel targets", logger.Attr("targets", targets))
 
 	if err := checkTargetLimit(len(targets), options); err != nil {
-		return nil, err
+		return err
 	}
 
-	var results []ecosystems.SCAResult
+	// processedFiles for the bazel resolver is computed at the resolver
+	// scope (WORKSPACE, MODULE.bazel, etc.), not per target. Attach to
+	// every emitted result; downstream consumers dedup.
+	processed := resolver.processedFiles()
+
+	emitted := 0
 	for _, target := range targets {
 		graph, err := resolver.buildDepGraph(ctx, target)
 		if err != nil {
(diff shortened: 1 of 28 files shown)


pkg/ecosystems/plugin_interface.go
```diff
@@ -15,23 +15,45 @@ type ResolverMetadata struct {
...
+// SCAResult{Descriptor: ..., Error: err} via onGraph — the run
+// continues so the caller sees every project the plugin attempted.
 type SCAPlugin interface {
-	BuildDepGraphsFromDir(ctx context.Context, log logger.Logger, dir string, options *SCAPluginOptions) (*PluginResult, error)
+	BuildDepGraphsFromDir(
+		ctx context.Context,
+		log logger.Logger,
+		dir string,
+		options *SCAPluginOptions,
+		onGraph func(SCAResult) error,
```
james-snyk: should we do `type OnGraphFunc func(SCAResult) error` given we have to repeate this function everywhere?

pkg/ecosystems/scatest/scatest.go
```diff
@@ -0,0 +1,34 @@
+// Package scatest provides shared helpers for SCAPlugin tests across
```
james-snyk, line 1: instead of just making this for tests might it be nicer if we included some pre defined funcs thats could be used

pkg/ecosystems/bazel/plugin.go
```diff
@@ -78,18 +84,18 @@ func (p Plugin) BuildDepGraphsFromDir(
...
-		})
+			ProcessedFiles: processed,
+		}); err != nil {
+			return err
+		}
+		emitted++
 	}
 
-	if len(results) == 0 {
+	if emitted == 0 {
```
tonidevine1, line 94: Bazel tracks an emitted counter here after the streaming migration, but the other plugins (Gradle, pip, pipenv etc.) don't have an equivalent. Was that intentional? Just checking whether Bazel needs special handling for the "ran but produced no successful graphs" case. Looks like it's just because that's how Bazel was set up prior to this PR but I'm just flagging the inconsistency as maybe we want to keep this unified?

pkg/ecosystems/scatest/scatest.go
cjheppell: I'm not hugely in favour of being opinionated on _how_ consumers of this interface should be handling the results as they come out. 

I'd much rather consumers think about the environment they're running in, and make a conscious decision on how to handle results as they arrive. That might be a slice, it might be a map, it might be streaming over a network, it might be writing to a filesystem. Since there's many different ways that a consumer wants to use these, I don't think exporting predefined funcs here would be hugely helpful.

I could be convinced if there's a strong argument though!

pkg/ecosystems/bazel/plugin.go
cjheppell: Yeah, so this is a symptom of how bazel was implemented. Since we no longer aggregate and return a slice, we have to keep track of how many we actually "emitted". This is different to other resolvers, but preserves existing behaviour of Bazel.

tonidevine1: Yeah make sense, thank you. I guess the question is should Bazel eventually emit `SCAResult{Error}` on per-target failures like the other plugins, so downstream sees partial failures?

tonidevine1 (approved):

pkg/ecosystems/plugin_interface.go
cjheppell: Yeah, fair enough. Pushed a new commit to create a func 👍

mcombuechen (approved):

PR #196, opened 2026-05-29
feat: [CMPA-542] Add NormalizeDepsPostHook to Gradle resolver

calhar-snyk:

### What this does

This adds the basic `NormalizeDepsPostHook` pattern for the Gradle resolver and adds a simple No-op hook.

### DIFF
--- a/pkg/ecosystems/gradle/plugin.go
+++ b/pkg/ecosystems/gradle/plugin.go
@@ -27,13 +27,23 @@ const (
 
 // Plugin is the Gradle SCA plugin.  It has no exported fields; all
 // configuration is passed through ecosystems.SCAPluginOptions at call time.
-type Plugin struct{}
+type Plugin struct {
+	normalizeDepsPostHook NormalizeDepsPostHook
+}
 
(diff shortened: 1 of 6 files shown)


pkg/ecosystems/gradle/plugin.go
```diff
@@ -36,6 +38,14 @@ func (p Plugin) GetName() string {
 	return PluginName
 }
 
+func NewGradlePlugin() *Plugin {
+	return &Plugin{normalizeDepsPostHook: nil}
+}
+
+func NewGradlePluginWithNormalizeDepsHook(normalizeDepsPostHook NormalizeDepsPostHook) *Plugin {
```
tonidevine1, line 45: Do we need both constructors? Could NewGradlePlugin() use NewNormalizeDepsPostHook() by default instead of nil? Wondering if I'm overlooking something here?

calhar-snyk: It's mostly just a syntactic nicety for the existing tests and later for isolated builds where we will _probably_ not want to use the hook.

tonidevine1 (approved):

PR #197, opened 2026-05-29
feat: [CMPA-542] Adds Gradle dependency normalization

calhar-snyk:

### What this does

Implements --gradle-normalize-deps flag that post-processes dependency graphs to use canonical Maven coordinates. Queries Snyk Packages API by SHA1 to resolve alternative artifact names/versions to their canonical forms, enabling consistent dependency identification across builds.

**Key features**:
- Extracts SHA1 checksums from dependency PURLs (requires --include-provenance)
- Queries Snyk Packages API to resolve canonical Maven GAV coordinates
- Rewrites dependency graphs with normalized names/versions while preserving SHA1 identity
- Deduplicates dependencies that resolve to the same canonical coordinates
- Handles suffixed nodes (constraint/pruned) correctly during rewriting
- Gracefully handles failures by leaving unmappable dependencies unchanged
- Concurrent API lookups with deduplication across multiple dependency graphs

**Usage**: --gradle-normalize-deps (requires authenticated org context)

**Implementation notes**:
- Post-processing hook architecture allows orthogonal feature composition
- Validates rewritten graphs using BuildGraph() to catch potential bugs
- Best-effort approach: individual lookup failures don't abort the entire scan

### DIFF
--- a/internal/snykclient/packages.go
+++ b/internal/snykclient/packages.go
@@ -1,0 +1,110 @@
+package snykclient
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+const packagesAPIVersion = "2024-10-15"
+
+// MavenPackageQuery describes the input to the Snyk Packages endpoint when
+// looking up the canonical Maven coordinates for an artifact identified by SHA1.
+type MavenPackageQuery struct {
+	Sha1     string
+	GroupID  string
+	Artifact string
+	Version  string
+}
+
+// packagesResponse mirrors the relevant subset of the /rest/packages response.
+type packagesResponse struct {
+	Data []struct {
+		ID string `json:"id"`
+	} `json:"data"`
+}
+
+// LookupMavenPackage resolves a Maven artifact's canonical coordinates by SHA1.
+//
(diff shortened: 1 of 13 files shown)


internal/snykclient/packages.go
```diff
@@ -0,0 +1,110 @@
...
+		return "", fmt.Errorf("failed to decode packages response: %w", err)
+	}
+	if len(body.Data) == 0 {
+		return "", nil
+	}
+	return body.Data[0].ID, nil
+}
+
+// buildPackagesAPIURL constructs the /rest/packages query URL. The endpoint is
+// org-agnostic at the path level; org context is provided via the API token.
```
tonidevine1, line 84: The comment says org context is provided via the API token, can you explain how that works for this endpoint?

calhar-snyk: So, some of our APIs want the org id either as part of the url itself (`https://api.snyk.io/rest/orgs/$orgId/packages/...` for example), or as part of a query parameter.
This one doesn't. It uses the auth token that it is provided with, and is built into the http client that is returned by `ctx.GetNetworkAccess().GetHttpClient()`, to determine the org id

tonidevine1 (approved):

PR #204, opened 2026-06-08
feat(npm): add lockfile-only resolver via npm ls

JamesPatrickGill:
## What this does

Adds an npm SCA plugin at `pkg/ecosystems/javascript/npmlocked/` that resolves dep graphs by shelling out to `npm ls` against the lockfile only — no `node_modules`, no `npm install`. Self-contained, not wired into the orchestrator in this PR. Supports `package-lock.json` and `npm-shrinkwrap.json` against npm 6 through current.

## Why

`snyk-nodejs-lockfile-parser`'s `npm-lock-v2` builder is ~700 lines of hand-rolled lockfile interpretation that we re-implement every time npm adds features. Delegating to `npm ls` shifts correctness onto npm and lets us inherit new features automatically.

## Customer-facing behavior alignments

This plugin matches the legacy parser's customer-visible behavior, with these deliberate changes:

- **Workspace optional deps included.** Legacy `--all-projects` hardcoded them off; single-project hardcoded them on. We include in both — these packages are in the lockfile because npm resolved and installed them.
- **Aliases use the underlying package name.** Snyk's vuln database keys on the published package identity, so the underlying name is what actually finds vulnerabilities. Customers with policies keyed on alias names will need to migrate.
- **Out-of-sync lockfiles produce partial graphs.** npm's `problems` array is surfaced via the logger; the scan succeeds with the partial tree instead of failing opaquely.
- **Missing optional deps are not synthesized as placeholders.** If the dep isn't in the lockfile, it isn't on disk — no fabricated nodes.
- **Bundled deps are currently invisible.** Open product question

## What's NOT in this PR

- Orchestrator registration — follow-up PR to avoid conflicting with other in-flight resolver work.
- 84-fixture legacy corpus + side-by-side divergence harness — PR #206.
- CircleCI multi-version matrix — PR #206.
- Bundled-dep recovery via lockfile post-pass — open product decision.

## How to verify

```sh
go test ./pkg/ecosystems/javascript/npmlocked/...
make test-npm-integration
make lint
```

## Risk

Not orchestrator-registered — no effect on production behavior. Reviewers can focus on plugin code in isolation.


JamesPatrickGill:
There will be further improvements for legacy comparison in a follow on ticket

PR #213, opened 2026-06-09
feat(pnpm): lockfile-only resolver + Rush adapter

jsutherland-snyk:
## What

Adds a pnpm SCA plugin at `pkg/ecosystems/javascript/pnpm/` that resolves pnpm dependency graphs by shelling out to `pnpm list --lockfile-only --json` — no `node_modules`, no install — plus a **Rush.js** adapter. Conforms to the established plugin pattern (bun/npm/yarn): `SCAPlugin` contract, runner interface + version gate, discovery 3-branch switch, one graph per importer with workspace members as stop-set leaves.

## How

- **pnpm** — `executor.go` shells out to `pnpm list --lockfile-only --json --depth Infinity`; `depgraph.go` builds one `godepgraph` per importer from the forward tree (`link:`/workspace deps rendered as sibling coordinates and treated as stop-set leaves; edge-dedup; cycle guard; version fallback).
- **Rush** (`rush.go`) — detects `rush.json`, gates on `pnpmVersion` (skips npm/yarn-backed and subspaces repos), and reconstructs the `common/temp` workspace context Rush generates at `rush install` time — in a **throwaway tmp tree** (lockfile copy + synthesized `pnpm-workspace.yaml` + copied member `package.json`s), never mutating the scanned repo — so pnpm resolves the committed `common/config/rush/pnpm-lock.yaml` on a clean checkout. A project listed in `rush.json` but missing its `package.json` is skipped-and-logged, not fatal.

## Tests

- Unit (default `go test`): `parsePnpmVersion`, `stripJSONComments`.
- Integration (`//go:build integration && pnpm`, `make test-pnpm-integration`) via `scatest` over `testdata/`: Rush workspace, standalone pnpm, missing-project skip, and npm/yarn/subspaces skip.

## Scope — NOT in this PR (deliberate)

- **Orchestrator registration is deferred.** The plugin is self-contained and not wired into `orchestrator/registry.go`. Registration behind a `FlagPnpmResolver` is the follow-up PR.
- Pre-GA follow-ups (from a PM review), to land with wiring:
  - **Dual-FF mutual exclusivity** — when `FlagPnpmResolver` is on, force the legacy `rushPnpmSupport` FF off for the same scan (the orchestrator runs the legacy CLI last with `--exclude-paths`; otherwise risk double-emission / `snyk monitor` identity churn in `--all-projects`).
  - **User-visible skip warnings** for npm/yarn/subspaces (currently logged only).
  - **Out-of-sync lockfile policy** (`--strict-out-of-sync`).
  - **Detection contract** for nested/sidecar pnpm projects under a Rush root.
- pnpm-feature parity (catalogs / aliases / `patch:`) is an independent-builder assumption to verify — this is **not** a reuse of the TS `nodejs-lockfile-parser`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/ecosystems/javascript/pnpm/depgraph.go
+++ b/pkg/ecosystems/javascript/pnpm/depgraph.go
@@ -127,10 +127,17 @@ func addDep(
 	visited, edgesSeen map[string]bool,
 ) error {
 	version := d.Version
+	// A workspace cross-dependency is identified by pnpm's authoritative `link:`
+	// version prefix — NOT by name. Keying on name alone would mis-flag a real
+	// registry package that happens to share a name with a workspace project,
+	// stop-set'ing it and silently dropping its subtree. Once a link is
+	// confirmed, consult wsVersions only to render the sibling's real version
+	// instead of the bare `link:<path>`.
 	isWorkspace := strings.HasPrefix(version, "link:")
-	if v, ok := wsVersions[name]; ok {
-		isWorkspace = true
-		version = v // render the sibling project's real version, not "link:.."
+	if isWorkspace {
+		if v, ok := wsVersions[name]; ok {
+			version = v
+		}
 	}
 	if version == "" {
 		version = defaultVersion
--- a/pkg/ecosystems/javascript/pnpm/depgraph.go
+++ b/pkg/ecosystems/javascript/pnpm/depgraph.go
@@ -1,0 +1,166 @@
+package pnpm
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	godepgraph "github.com/snyk/dep-graph/go/pkg/depgraph"
+)
+
+const pkgManager = "pnpm"
+
+// depGraphResult pairs a built dep graph with the path to the package.json that
+// describes it, relative to the scan root.
+type depGraphResult struct {
+	graph          *godepgraph.DepGraph
+	pkgJSONRelPath string
+}
+
+// buildDepGraphs produces one dep graph per importer project from `pnpm list`
+// output. Other workspace members appear as stop-set leaves in every graph
+// (their subtrees are walked only in their own graph), preventing duplicated
+// reporting across workspaces. The importer whose directory is skipDir (the
+// Rush "rush-common" aggregate at the staged common/temp) is omitted; pass ""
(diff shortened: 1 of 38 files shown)


pkg/ecosystems/javascript/pnpm/depgraph.go
```diff
@@ -0,0 +1,166 @@
...
+	d listDep,
+	wsVersions map[string]string,
+	visited, edgesSeen map[string]bool,
+) error {
+	version := d.Version
+	isWorkspace := strings.HasPrefix(version, "link:")
+	if v, ok := wsVersions[name]; ok {
+		isWorkspace = true
+		version = v // render the sibling project's real version, not "link:.."
+	}
```
JamesPatrickGill: This name check means we key wsVersions on name only right? So if we manage to collide a workspace name with a real name on the public registry we will have silent skips.

It's an edgecase we have seen before

JamesPatrickGill (commented):
I think this is fine, but it does seem code wise to treat rush as a first class citizen and pnpm as second. Could we consolidate the logic to have them more similar - plugin is seeming bloated? I think making them the same resolver is okay as long as rush stays as a PNPM wrapper

pkg/ecosystems/javascript/pnpm/depgraph.go
jsutherland-snyk: Fixed in 24c6571. Workspace deps are now identified by pnpm's authoritative `link:` version prefix rather than name membership in `wsVersions`, so a transitive registry package that happens to share a name with a workspace project is no longer mis-flagged as a workspace leaf / stop-set — its subtree gets walked normally. `wsVersions[name]` is consulted only to render the sibling's real version once a `link:` is confirmed. Added a regression test (`TestBuildDepGraphs_NameCollisionIsNotStopSet`, plus `_WorkspaceLinkResolvesSiblingVersion` for the genuine link case).

Also passed `scanTarget` by pointer in `runAndBuild` (`&targets[i]`) to clear a `gocritic` hugeParam the scanTarget refactor introduced — lint is green now. Thanks for the catch!

JamesPatrickGill (approved):

PR #222, opened 2026-06-22
feat: [CMPA-604] Adds support for passing `--include-component-metadata` flag to legacy cli

calhar-snyk:

### What this does

This adds support for a new legacy CLI command flag `--include-component-metadata`, which will be passed down from `cli-extension-sbom`.

The flag causes the downstream dependency graph output to include hash and distribution url information as labels on the DepGraph, which can then be used to populate components[].hashes[] entries and provide a distribution entry for externalUrls in the CycloneDX format SBOM.

This is currently going to be implemented for the legacy mvn and npm plugin paths.

### DIFF
--- a/internal/legacycli/legacy_resolution.go
+++ b/internal/legacycli/legacy_resolution.go
@@ -274,6 +274,11 @@ func PrepareLegacyFlags(argument string, cfg configuration.Configuration, logger
 		logger.Print("Include provenance: true")
 	}
 
+	if cfg.GetBool(workflow.FlagIncludeComponentMetadata) {
+		cmdArgs = append(cmdArgs, "--include-component-metadata")
+		logger.Print("Include component metadata: true")
+	}
+
 	if cfg.IsSet(workflow.FlagDotnetRuntimeResolution) {
 		dotnetRuntimeResolution := cfg.GetBool(workflow.FlagDotnetRuntimeResolution)
 		cmdArgs = append(cmdArgs, fmt.Sprintf("--dotnet-runtime-resolution=%t", dotnetRuntimeResolution))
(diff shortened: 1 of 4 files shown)


james-snyk (approved):

danielroymoore (approved):

internal/workflow/flags.go
```diff
@@ -30,6 +30,7 @@ const (
 	FlagNugetPkgsFolder               = "packages-folder"
 	FlagUnmanagedMaxDepth             = "max-depth"
 	FlagIncludeProvenance             = "include-provenance"
+	FlagIncludeHashes                 = "include-hashes"
```
snyk-tim: The PR title says `--include-component-metadata` but this defines the flag as `--include-hashes`.

Is the `--include-hashes` ecosystem specific or is it a generic flag used across ecosystems?

internal/legacycli/legacy_resolution.go
```diff
@@ -274,6 +274,11 @@ func PrepareLegacyFlags(argument string, cfg configuration.Configuration, logger
 		logger.Print("Include provenance: true")
 	}
 
+	if cfg.GetBool(workflow.FlagIncludeHashes) {
+		cmdArgs = append(cmdArgs, "--include-hashes")
```
snyk-tim: It looks like this should be using `--include-component-metadata` which matches that changes made to [snyk-maven-plugin](https://github.com/snyk/snyk-mvn-plugin/blob/7ded31906c4e329536dc5bceabefd59d48e565e0/lib/index.ts#L176)

snyk-tim (approved):

internal/workflow/flags.go
snyk-tim: Just to follow up, it is confirmed that `--include-component-metadata` will be used in `npm` also.

